PR #1, opened 2022-06-22
build: init library

peterpeterparker:
# Motivation

Init the build and tools we need for the design system.

# Changes

- setup `SvelteKit`
- setup `Storybook`
- add `jest` and `testinglibrary` for test purpose
- add `CODEOWNERS` and PR template for GitHub


src/stories/Button.stories.js
```diff
@@ -0,0 +1,51 @@
+import Button from "./Button.svelte";
```
mstrasinskis, line 1: The empty Button.svelte looks suspicious but not critical for the PR.

mstrasinskis (approved):
LGTM, nice base to start. Thank you.

src/stories/Button.stories.js
peterpeterparker: Good point, removed it. There are enough other examples.

PR #7, opened 2022-07-24
feat: layout, menu and toolbar components + docs

peterpeterparker:
# Motivation

This PR introduces layout, menu and toolbar components. While almost all components where copied from NNS-dapp, `Layout` is a new component that just encapsulate the existing layout to make the use of the library easier to setup.

In addition, this PR also introduces a change in the concept for the showcase. Instead of using Storybook, I suggest to use SvelteKit to build a custom dapp that showcase and document the design system.

According a feedback we had in a recent interview, Storybook might become a bit slow while the design grow and is also slower in term of dev experience (e.g. does not use vite). While it has other advantages, a custom website gives us also flexibility.

It's also quite meta to use the design system to document the design system 😉.

# Note

The docs website still need some styling. This PR just sets bare minimum styling.

If the idea also works for you @mstrasinskis, I'll drop Storybook in another PR.

# Changes

1. Component library:

- copy `Backdrop`, `Menu`, `MenuButton`, `MenuItem`, `SplitPane` and `Toolbar` from NNS-dapp
- add i18n for the library and docs the fact that it is currently static en
- add `Layout` component that replicates existing layout
- configure build and jest to use `$lib` path alias
- docs layout
- build bump dependencies

2. Docs website

- build with `mdsvex` as well to support markdown ("mdx for svelte")
- open some pages and related menu links
- add some styling for the docs website only

# Screenshots

Capture d’écran 2022-07-24 à 10 59 50
Capture d’écran 2022-07-24 à 10 59 54


src/docs/components/DocsMenu.svelte
```diff
@@ -0,0 +1,24 @@
...
+  } = $page);
+</script>
+
+<MenuItem href="/" selected={pathname === "/"} on:click>Home</MenuItem>
+<MenuItem href="/start" selected={pathname === "/start"} on:click>
+  Getting Started
+</MenuItem>
+<MenuItem
+  href="/components"
+  selected={pathname.indexOf("/components") > -1}
```
mstrasinskis, line 17: startsWith?

mstrasinskis (approved):
LGTM, thank you!

src/docs/components/DocsMenu.svelte
peterpeterparker: Nice

PR #10, opened 2022-08-01
feat: Card, Back and docs improvements

peterpeterparker:
# Motivation

- Provide components `Card`, `CardGrid`  and `Back` with the design system.
- Improve documentation content and style

# Changes

Design system:

- cmp `Card`, `CardGrid`  and `Back` copied from NNS-dapp
- related icons and i18n copied as well
- fix back button display in layout (it needs a property as the detection of provided slot is static)

Docs dapp:

- use above component and add cards on welcome screen
- docs "principles"and example of theming
- implement back button for sub routes
- some custom styling notably to set some spacing between paragraphs
- update README to avoid to double documentation texts

# Screenshots

Capture d’écran 2022-08-01 à 15 51 33
Capture d’écran 2022-08-01 à 15 51 40
Capture d’écran 2022-08-01 à 15 51 43


mstrasinskis (approved):
LGTM 👍

PR #20, opened 2022-08-08
feat: toolbar end and back button

peterpeterparker:
# Motivation

Components:

- The layout should redirect the toolbar `end` slot  so that the consumer can use it - e.g. to display a user button
- Simplify the usage of the back action. Instead of passing a slot, add the `<Back />` component in the layout

Docs:
- While developing this noticed there was an issue in the `docs` website regarding the back button

# Changes

Components:

- add a `slot=toolbar-end` in the layout that target the `toolbar=end` (exposing "end" felt a bit generic so I gave it another name in the layout)
- include back button in layout
- modify docs accordingly

Docs:

- fix back action in the docs website

### DIFF
--- a/src/lib/components/Layout.svelte
+++ b/src/lib/components/Layout.svelte
@@ -24,7 +24,7 @@
 
       <slot name="title" />
 
-      <slot name="more" slot="end" />
+      <slot name="toolbar-end" slot="end" />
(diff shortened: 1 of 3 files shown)


mstrasinskis (approved):
LGTM

PR #22, opened 2022-08-09
feat: support for deprecated layout

peterpeterparker:
# Motivation

For backwards compatibility, the design system needs to support the old layout as well.

# Changes

- add property `modern=true` to `<Layout />` and unset padding and margin on `main` element if old layout


mstrasinskis (approved):
LGTM

PR #23, opened 2022-08-09
feat: expose icons

peterpeterparker:
# Motivation

Some might needs the icons while using this library, notably the "close icon".

# Changes

Lib:

- export icons to consumers

Docs:

- list icons at build time and add a dedicated display page


mstrasinskis (approved):
LGTM. Thanks

PR #25, opened 2022-08-09
refactor: move assets to subfolder

peterpeterparker:
# Motivation

NNS-dapp loads fonts and other static content from an `assets` folder. As, if I am not incorrect, CSS `url()` does not support variables, we have to set the same path to the fonts in this library.

# Changes

- move assets to sub-folder `assets`
- adapt fonts path


mstrasinskis (approved):
LGTM 👍

PR #26, opened 2022-08-09
build: package first and then update version

peterpeterparker:
# Motivation

The modified version is overwritten by the package step. We should first package and then update the version number. This is different than in nns-dapp or ic-js.


mstrasinskis (approved):
LGTM

PR #28, opened 2022-08-09
build: fix npm missing user

peterpeterparker:
# Motivation

The publish to npm is failing (see https://github.com/dfinity/gix-components/runs/7750678770?check_suite_focus=true). Most probably because the token is set to the root and not the working directory `package`.

# Changes

- set `npmrc` in working dir


mstrasinskis (approved):
LGTM

PR #30, opened 2022-08-10
feat: deprecate <CardGrid /> for CSS

peterpeterparker:
# Motivation

The grid might be another type of elements than a `div`. We notably need the grid to support `InfiniteScroll` too which happens to be a `ul`. For simplicity reason and to promote the usage of CSS rather than JavaScript properties, this PR deprecates `<CardGrid />` and introduces a new global style `.grid`

# Changes

Lib:

- remove `<CardGrid />`
- refactor style to global style `grid.scss`

Docs:

- dedicated grid docs


mstrasinskis (approved):
LGTM. Very well documented, thanks!

PR #31, opened 2022-08-10
feat: infinite scroll

peterpeterparker:
# Motivation

Move infinite scroll component from NNS-dapp.

# Changes

- lib: new component `<InfiniteScroll />` copied from NNS-dapp
- lib: fix `main` height to fit the all content
- docs: add page to documet components


src/docs/styles/docs-global.scss
```diff
@@ -63,6 +64,10 @@ p code {
   font-size: inherit;
 }
 
+p strong {
+  color: #ffffff;
```
mstrasinskis, line 68: Will the docs support theming in the future?

mstrasinskis (approved):
LGTM 👍

peterpeterparker:
@mstrasinskis by testing this PR I also noticed that the `main` element was not fitting well the height of its parent so I modified to `fit-content` and integrated it here.

src/docs/styles/docs-global.scss
peterpeterparker: good point, I updated it with the contrasting color of the background:

```
p strong {
  color: var(--background-contrast);
}
```

PR #34, opened 2022-08-15
feat: infinite scroll grid option

peterpeterparker:
# Motivation

In the new ui we will have to use the infinite scroll as grid too.

# Changes

- add a `layout` property to `<InfiniteScroll />` component (`list` or `grid`)


mstrasinskis (approved):
LGTM. So convenient with the class based layout 👍

PR #35, opened 2022-08-15
fix: class layout

peterpeterparker:
# Motivation

The class should receive the value of the prop.

# Changes

- `class:layout` renders `class="layout" we want `class={layout}` to render the value


mstrasinskis (approved):
LGTM

PR #36, opened 2022-08-16
style: fix card height in grid

peterpeterparker:
# Motivation

The selector was valid for infinite scroll grid but not normal grid.

# Changes

- update card selector in grid to apply margin and height

# Screenshots

is:

Capture d’écran 2022-08-16 à 07 44 32

fixed:

Capture d’écran 2022-08-16 à 07 44 48


mstrasinskis (approved):
LGTM 👍

PR #37, opened 2022-08-16
feat: header title component

peterpeterparker:
# Motivation

An opinionated component to display a title in the header of the dapp.

# Note

Style refactored from nns-dapp `header.scss`

# Changes

- lib: new component `<HeaderTitle />`
- docs: usage in layout and related documentation


src/lib/components/Title.svelte
```diff
@@ -0,0 +1,15 @@
+<h4><slot /></h4>
+
+<style lang="scss">
+  @use "../styles/mixins/text";
+
+  h4 {
+    margin: 0;
+    color: white;
```
mstrasinskis, line 8: This is where having value/reference tokens would be handy.

src/lib/components/Title.svelte
```diff
@@ -0,0 +1,15 @@
+<h4><slot /></h4>
```
mstrasinskis, line 1: Probably `Title` is a bit a too generic name for so specific component but not a strong opinion.

mstrasinskis (approved):
LGTM. Thank you.

src/lib/components/Title.svelte
peterpeterparker: yes definitely!

note: style is the same as before except a grid position attribute which was actually unused
it disappeared because the component was removed in nns-dapp. I copied the style from history, no other change.

src/lib/components/Title.svelte
peterpeterparker: Not super happy with the name neither, better idea?

`HeaderTitle`?

peterpeterparker: Done as discussed renamed to `HeaderTitle`

PR #38, opened 2022-08-16
fix: layout legacy footer

peterpeterparker:
# Motivation

The footer of the legacy layout is broken - i.e. is not sticky anymore.

# Changes

- defer the responsibility to the consumer to use the `main` and `footer` elements
- add a `scrollable-content` to the split pane (otherwise footers position absolute won't be sticky)
- rename `nns` option in `legacy`

### DIFF
--- a/src/lib/styles/global/layout.scss
+++ b/src/lib/styles/global/layout.scss
@@ -13,16 +13,6 @@ main {
   @include media.min-width(medium) {
     padding: var(--padding-4x);
   }
-
-  &.legacy {
-    margin: inherit;
-    padding: inherit;
-    max-width: inherit;
-
-    @include media.min-width(medium) {
-      padding: inherit;
-    }
-  }
 }
 
 header {
--- a/src/lib/styles/global/layout.scss
+++ b/src/lib/styles/global/layout.scss
@@ -4,12 +4,24 @@ main {
   height: fit-content;
   width: 100%;
 
+  margin: 0 auto;
+
   padding: var(--padding-4x) var(--padding-2x);
   max-width: var(--max-main-content-width);
   box-sizing: border-box;
 
   @include media.min-width(medium) {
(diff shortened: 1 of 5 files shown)


src/lib/styles/global/layout.scss
```diff
@@ -32,3 +44,10 @@ header {
 
   --toolbar-pointer-events: all;
 }
+
+footer {
```
lmuntaner, line 48: to be honest, I don't think global styles is the way to go for a components library. Shouldn't the library just expose components and maybe a theme that the components use?

Not for this PR though...

peterpeterparker: it's like for the [grid](https://ipi2f-uqaaa-aaaad-aabza-cai.ic0.app/components/grid) or the button - when it does not need JavaScript, I think it's more performant to not use JavaScript. Being said, for sure it would need design tokens as they are currently doing in II, that would be way cleaner! Also docs improvements.

src/lib/styles/global/layout.scss
```diff
@@ -4,13 +4,25 @@ main {
...
+
   padding: var(--padding-4x) var(--padding-2x);
   max-width: var(--max-main-content-width);
   box-sizing: border-box;
 
   @include media.min-width(medium) {
     padding: var(--padding-4x);
   }
+
+  &.legacy {
```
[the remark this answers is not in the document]
peterpeterparker: Yes sure definitely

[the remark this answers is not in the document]
peterpeterparker: Done

src/lib/styles/global/layout.scss
peterpeterparker: but you are right, there is a bit a mix though, probably worth a big clean up

PR #40, opened 2022-08-20
build: sveltekit breaking changes

peterpeterparker:
# Motivation

SvelteKit introduced a new syntax for its routing, as well as for its data loading, in the newest pre-1.0 release.
This PR performs the migration.

# Resources

- Migration guide: https://github.com/sveltejs/kit/discussions/5774
- Tutorial: https://joshcollinsworth.com/blog/sveltekit-breaking-changes

# Changes

- bump all dependencies
- migrate routes
- migrate package script
- adapt configuration


peterpeterparker:
Blocked. The [certified assets canister](https://github.com/dfinity/certified-assets/) does not support SvelteKit routing anymore.

src/routes/+layout.svelte
```diff
@@ -3,7 +3,7 @@
   import DocsMenu from "$docs/components/DocsMenu.svelte";
   import { page } from "$app/stores";
   import { goto, afterNavigate } from "$app/navigation";
-  import HeaderTitle from "../lib/components/HeaderTitle.svelte";
```
lmuntaner, line 6: oh, is this some new cool svelte magic?

vite.config.js
```diff
@@ -7,7 +7,6 @@ const config = {
   resolve: {
     alias: {
       $docs: resolve("./src/docs"),
-      $lib: resolve("./src/lib"),
```
lmuntaner, line 10: why is one removed and the other stays?

lmuntaner (approved):
A couple of comments, already approving.

src/routes/+layout.svelte
peterpeterparker: you mean the `$lib`? not new, I think you can do that with rollup or any bundler - i.e. we can do that in NNS-dapp too if we want too but we would need an external lib. In SvelteKit it's included and configured out of the box.

vite.config.js
peterpeterparker: before last version of SvelteKit/Vite we had to overwrite the default `$lib` to be able to declare `$docs` as well but, after the migration, it's the contrary, `$lib` should not be re-declared as available in any case. I guess the custom config used to overwrite the default config while now they merge it.

lmuntaner: ok, thanks for the explanation!

src/routes/+layout.svelte
lmuntaner: ah, ok, let's think about it. For me it would depend more on the VS Code integration to be honest.

PR #41, opened 2022-08-22
feat: two columns & two rows, two columns

peterpeterparker:
# Motivation

Move `<TwoColumns />` cmp from NNS-dapp and add a new component for two rows and two columns.

# Changes

- lib: copy - move - component `<TwoColumns />` cmp from NNS-dapp
- lib: rename `left` and `right` slots to `start` and `end`
- lib: extract flex and grid to mixin
- lib: create new `<Content />` cmp to display two columns and two rows ("ACBD"")
- docs: brief documentation for `<TwoColumns />`

### DIFF
--- a/src/lib/components/Content.svelte
+++ b/src/lib/components/Content.svelte
@@ -1,0 +1,70 @@
+<script lang="ts">
+  export let testId: string | undefined = undefined;
+
+  let itemA: boolean;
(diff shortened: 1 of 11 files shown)


peterpeterparker:
We close this PR to follow a CSS based approach as in #42 - "component only when js needed". 
Discussed with @mstrasinskis

PR #45, opened 2022-08-26
feat: infinite scroll observe last element

peterpeterparker:
# Motivation

Make the `<InfiniteScroll />` component compatible for both list and grid. To do so, update behavior to observe last element of the container.

# Changes

- observe last element of the container instead of calculated position
- remove unused constants
- `disconnect` on `beforeUpdate` instead of `afterUpdate` to prevent the component to emitting on layout shift
- add cmp property `disabled` to disable observation ("to disable the component when everything is fetched")

### DIFF
--- a/src/lib/components/InfiniteScroll.svelte
+++ b/src/lib/components/InfiniteScroll.svelte
@@ -50,7 +50,7 @@
     // The DOM has been updated. We reset the observer to the current last HTML element of the infinite list.
 
     // If not children, no element to observe
-    if (container.children.length === 0) {
+    if (container.lastElementChild === null) {
       return;
     }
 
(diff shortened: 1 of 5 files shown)


src/lib/components/InfiniteScroll.svelte
```diff
@@ -42,57 +43,36 @@
...
-    observer.observe(element);
+    observer.observe(container.lastElementChild);
   });
 
   onDestroy(() => observer.disconnect());
+
+  // The infinite scroll "disabled" property is updated. If it becomes "false", the observer should be disconnected.
+  $: disabled,
+    () => {
+      if (!disabled) {
```
mstrasinskis, line 70: Isn't this covered by `beforeUpdate` callback?

mstrasinskis (approved):
LGTM, just one question.

src/lib/components/InfiniteScroll.svelte
peterpeterparker: You are right I think, removed that auto-subscriber. Thanks!

PR #48, opened 2022-09-01
style: line color and cell gap

peterpeterparker:
# Motivation

Mischa's review.

# Changes

- adjust cell gap spacing
- introduce line color

### DIFF
--- a/src/lib/styles/themes/dark.scss
+++ b/src/lib/styles/themes/dark.scss
@@ -63,6 +63,10 @@
   --disable: #505050;
   --disable-contrast: #969696;
 
+  // Line
+  --line: #505050;
+  --line-contrast: #969696;
+
   // Palette
   --background: #383c3c;
   --background-rgb: 56, 60, 60;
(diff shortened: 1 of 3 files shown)


src/lib/styles/themes/dark.scss
```diff
@@ -63,6 +63,10 @@
   --disable: #505050;
   --disable-contrast: #969696;
 
+  // Line
+  --line: #505050;
```
mstrasinskis, line 67: Out of curiosity, where are we gonna use it?

mstrasinskis (approved):
LGTM.

src/lib/styles/themes/dark.scss
peterpeterparker: Mischa said "there are lots of lines in the design" and also that he use the term "line" in his Figma file.

Currently we use the line in the neuron of the proposal detail and we will use also the color as background of the detail (see screenshot in red)
Capture d’écran 2022-09-02 à 12 45 49

mstrasinskis: Thank you for the update.

PR #49, opened 2022-09-01
feat: skeleton text

peterpeterparker:
# Motivation

Add the skeleton component to gix-components.

# Note

The component was copied from NNS-dapp but instead of rendering a sole `span` it has been upgraded to support different types of HTML elements. This will be useful for the new proposal detail page.

# Changes

- lib: copy NNS-dapp `<SkeletonParagraph />` to `<SkeletonText />`
- lib: add option to render `span`, `p` or `h1-3`
- docs: add new doc page to document option and showcase


mstrasinskis (approved):
LGTM, thanks

PR #51, opened 2022-09-02
Icons copied to gix-components

wackyleo459:
# Motivation

All current Icons from NNS-Dapp library transferred to gix-components

# Changes

- layout for grid customized for better spacing
 - Sizing added for some icons

# Screenshots

Before:
Due to name being too long, the view looks crowded

Screen Shot 2022-09-02 at 3 37 14 PM

After:
Grid view looks better
Screen Shot 2022-09-06 at 11 06 31 PM


peterpeterparker:
It's actually probably easier to stick to the current names I would say. Doing so, when we will replace the icons in NNS-dapp we will just have to modify the `import` while if we modify the names as well, we would have to modify `import` and `component names`.

In addition, names inherit the original name of the open source icons we are using.

For these two reasons, I would suggest to stick to the current name.

Regarding the documentation showcase, you are right, it looks crowded. We can improve this with CSS in [icons.svelte](https://github.com/dfinity/gix-components/blob/main/src/routes/icons.svelte)

peterpeterparker:
You might want to update the branch and `npm ci`, I have migrated SvelteKit i main which contains quite some changes in the routing.

wackyleo459:
@peterpeterparker Hi David, please review!

peterpeterparker (approved):

peterpeterparker:
Thanks! I merged the PR for you @wackyleo459 and deployed a new version of [gix.design](https://gix.design). Looks cool having all these icons.

I also published a new version to npm so that you can use the lib in NNS-dapp and replace the icons there 😃

peterpeterparker:
@wackyleo459 actually something is missing, wait for a PR...

peterpeterparker:
@wackyleo459 I created a PR #54 to expose the icons. We will need this to use these in NNS-dapp.

PR #55, opened 2022-09-09
feat: content height and scroll length with bottom sheet

peterpeterparker:
# Motivation

Adapt content height and scroll length if a bottom sheet is presented.

# Changes

- add a new store for the layout bottom offset
- update offset according window width and bottom sheet height in component
- add padding bottom to content

# Screenshot

Capture d’écran 2022-09-09 à 07 37 31
Capture d’écran 2022-09-09 à 07 37 36

### DIFF
--- a/src/lib/components/BottomSheet.svelte
+++ b/src/lib/components/BottomSheet.svelte
@@ -1,4 +1,15 @@
-<div role="dialog" data-tid="bottom-sheet">
+<script lang="ts">
+  import { layoutBottomOffset } from "../stores/layout.store";
+  import { onDestroy } from "svelte";
+
+  onDestroy(() => ($layoutBottomOffset = 0));
+</script>
+
+<div
+  role="dialog"
+  data-tid="bottom-sheet"
+  bind:clientHeight={$layoutBottomOffset}
+>
   <slot />
 </div>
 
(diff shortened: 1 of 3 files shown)


src/lib/components/BottomSheet.svelte
```diff
@@ -1,4 +1,21 @@
-<div role="dialog" data-tid="bottom-sheet">
+<script lang="ts">
+  import { layoutBottomOffset } from "../stores/layout.store";
+  import { onDestroy } from "svelte";
+
+  onDestroy(() => ($layoutBottomOffset = 0));
+
+  // 1024 is the $breakpoint-large size. See also the CSS code of this component. On large screen the bottom sheet is not sticky.
+  const updateBottomOffset = () =>
+    ($layoutBottomOffset = innerWidth < 1024 ? height : 0);
```
mstrasinskis, line 9: What do you think, does it make sense to expose breakpoints and read the value directly from the DOM (e.g. `getComputedStyle(document.documentElement).getPropertyValue("--breakpoint-large");`)?

mstrasinskis (approved):
LGTM. One question.

src/lib/components/BottomSheet.svelte
peterpeterparker: @mstrasinskis I would absolutely love that! Don't like to have that value hardcoded.

Can you take care of it in another PR?

PR #61, opened 2022-09-13
fix: typo in icon

peterpeterparker:
# Motivation

Fix icon in typo. "Footer" should not be there.


wackyleo459 (approved):
Ah, good to catch that typo!

PR #62, opened 2022-09-13
feat: update function to toastStore

mstrasinskis:
# Motivation

To be able to update the presented toast text, `update` function was added (important for the voting status update).

# Changes

- toastStore.`update`


peterpeterparker (approved):
Thanks

PR #64, opened 2022-09-14
style: bottom sheet safe area

peterpeterparker:
# Motivation

Respect safe area of iOS bottom inset for bottom sheet.

### DIFF
--- a/src/lib/components/BottomSheet.svelte
+++ b/src/lib/components/BottomSheet.svelte
@@ -33,6 +33,8 @@
 
     z-index: var(--bottom-sheet-z-index);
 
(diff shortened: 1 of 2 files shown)


src/lib/components/BottomSheet.svelte
```diff
@@ -33,13 +33,17 @@
 
     z-index: var(--bottom-sheet-z-index);
 
+    padding-bottom: env(safe-area-inset-bottom);
```
mstrasinskis, line 36: TIL, thanks

mstrasinskis (approved):
LGTM. Thank you.

PR #68, opened 2022-09-21
feat: in case of errors display toasts on top

peterpeterparker:
# Motivation

In case of errors display the toasts always on the top including above modals.

# Changes

- split `z-index` for toasts in two, one for normal behavior one for errors

# Screenshots
Capture d’écran 2022-09-21 à 08 30 18

Capture d’écran 2022-09-21 à 08 30 10

### DIFF
--- a/src/lib/components/Toasts.svelte
+++ b/src/lib/components/Toasts.svelte
@@ -2,11 +2,17 @@
   import { toastsStore } from "$lib/stores/toasts.store";
   import Toast from "./Toast.svelte";
   import { layoutBottomOffset } from "$lib/stores/layout.store";
+
+  let hasErrors: boolean;
+  $: hasErrors = $toastsStore?.find(({ level }) =>
+    ["error", "warn"].includes(level)
+  );
 </script>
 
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
Thanks

peterpeterparker:
My thought is that error and warning are important for users while the other are informative which sounds like a generic rule.

Also in NNS-dapp voting, vote in progress are "info" toast.

lmuntaner:
why would that info go behind the Modal? I don't see the advantage on hiding them.

Moreover, in mobile that means that they won't even be slightly visible.

peterpeterparker:
test the new voting tab, you will get it.

lmuntaner:
Sorry, I don't get it. Do you mean that the info toast is ok behind the Modal? As a user I would prefer that to be readable.

https://user-images.githubusercontent.com/4550653/191435825-418f0100-8d73-4741-a007-39ab5a86a77d.mov

peterpeterparker:
Yes exactly that's the goal. Test on mobile.

PR #70, opened 2022-09-21
style: review buttons style according figma design

peterpeterparker:
# Motivation

Review buttons style according figma design.

# Changes

- lib: buttons padding, border, hover, focus etc. according figma
- lib: `disabled` is the same for all buttons
- lib: `.small` button style deprecated
- docs: update buttons showcase with more flavors


mstrasinskis (approved):
LGTM, thanks

PR #71, opened 2022-09-22
fix: missing shadow

mstrasinskis:
# Motivation

Because of no z-index on the header the shadow is not visible. The shadow was moved to the `::after` block.

# Changes

- move box shadow to the `after` component

# Screenshots

image


peterpeterparker (approved):
Thanks 👍

PR #72, opened 2022-09-27
feat: modal toolbar sticky only for alert

peterpeterparker:
# Motivation

Discussed with Mischa, the modal's toolbar should not be sticky in case of dialog.

In case of alert, we keep a sticky option for backwards compatibility and time being.

# Changes

- rename `toolbar` slot to `footer` for "alert" (breaking change will need a pr in nns-dapp)
- add utility class `.toolbar` for "dialog"


mstrasinskis (approved):
LGTM, thanks.

PR #75, opened 2022-09-28
feat: input and range

peterpeterparker:
# Motivation

- Add `<Input />` and `<InputRange />` to `gix-components`
- Review input range design to be able to use it in dark mode with new design

# Changes

lib:

- move `<Input />` and `<InputRange />` from NNS-dapp
- adapt input to make it agnostic
- review range style
- rename CSS variable for input box shadow (will have to be adapted in NNS-dapp)

doc:

- add new related pages

### DIFF
--- a/src/tests/components/InputValueTest.svelte
+++ b/src/tests/components/InputValueTest.svelte
@@ -1,0 +1,24 @@
+<script lang="ts">
+  import { createEventDispatcher } from "svelte/internal";
+
+  import Input from "$lib/components/Input.svelte";
+
+  const dispatch = createEventDispatcher();
+
+  export let inputType: "text" | "icp" = "text";
+  export let name: string;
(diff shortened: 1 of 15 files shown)


src/tests/components/InputValueTest.svelte
```diff
@@ -0,0 +1,24 @@
...
+
+  export let inputType: "text" | "icp" = "text";
+  export let name: string;
+  export let value: string | undefined = undefined;
+  export let placeholder = "test.placeholder";
+
+  let amount: string | undefined = value;
+  $: amount, (() => dispatch("testAmount", { amount }))();
+
+  // We want to test that we can change value programtically
```
mstrasinskis, line 16:
```suggestion
  // We want to test that we can change value programmatically
```

mstrasinskis (approved):
LGTM, minor typo

PR #77, opened 2022-09-28
style: fix modal height

peterpeterparker:
# Motivation

We don't have it yet with the new modal but I guess we will.

# Changes

- fix modal height

# Screenshots

Capture d’écran 2022-09-28 à 15 59 02


mstrasinskis (approved):
LGTM, thanks

PR #79, opened 2022-10-05
Components/popover

wackyleo459:
# Changes
Popover component
- debounce utility function replicated
- a few icons changed to customize size for demonstration purpose

# Screenshots
Screen Shot 2022-10-05 at 3 33 57 PM

# Suggestion for next step
For Icons: Most of the icons (if not all) might be better suited with customizable size.


src/lib/utils/debounce.ts
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable-next-line @typescript-eslint/ban-types */
+export const debounce = (func: Function, timeout?: number) => {
```
peterpeterparker, line 2: can you add a comment about what we discussed, something like

```
// Duplicate NNS-dapp similar function for simplicity reason
```

src/routes/components/popover/+page.md
```diff
@@ -0,0 +1,141 @@
...
+    let selected1: boolean = true;
+</script>
+
+# Popover
+
+A popover is a floating container that is rendered over the content
+next to an anchor — commonly a button — which initiates its display.
+<br />
+
+# Showcase
```
peterpeterparker, line 19: for consistency reason with the other pages, can you move the showcase at the end of the page?

src/routes/components/popover/+page.md
```diff
@@ -0,0 +1,141 @@
...
+                <IconWallet size={34}/>
+            </span>Wallet
+        </div>
+        <div class="account">
+            <span class="account_icon">
+                <IconExit size={34}/>
+            </span>Logout
+        </div>
+    </Popover>
+    <button
```
peterpeterparker, line 57: you can simplify and remove the showcase "Apparence".

Once this PR added, I'll take care of adding the global dark/light mode theme switcher and capatibility to the gix-components.

peterpeterparker (commented):
Nice 👍

src/routes/components/popover/+page.md
wackyleo459: @peterpeterparker so the reason I moved the showcase to the top is because if it's at the bottom, the popover will overflow viewport since its placement is always below the anchor. Can we change the popover to account for that, if it's at the bottom? Otherwise the popover will be cut off when you view it.
Screen Shot 2022-10-06 at 11 16 06 AM

peterpeterparker: Gotcha, you are right, makes sense 👍

peterpeterparker (approved):
LGTM, thanks!

peterpeterparker (approved):
LGTM, thanks

PR #80, opened 2022-10-06
style: fix input inner box-shadow

peterpeterparker:
# Motivation

The CSS variable for the box-shadow of input has been renamed to be develop input range as in Figma but the input wasn't updated accordingly.

# Changes

- use correct css variable


mstrasinskis (approved):
LGTM

PR #81, opened 2022-10-07
feat: theme, theme toggle and theme

peterpeterparker:
# Motivation

Integrate the theme switcher, the theme toggle component and toggle component into the library.

# Note

The `theme.store` code moved is not the last version of NNS-dapp but previous one because we don't need the generic `writableStore` in the library. The store and related `theme.utils` were adapted to support the format of the property saved in `localStorage` - i.e. not a `string` but a `JSON.stringify(theme)` as currently on `main` branch of NNS-dapp.

# Changes

lib

- new `theme.store` and related utils (not exposed)
- component `<ThemeToggle />` and `<Toggle />` moved in the library

docs:

- related components pages added
- `Theming` doc updated
- add theme switcher to docs header with a popover


src/lib/utils/enum.utils.ts
```diff
@@ -0,0 +1,7 @@
+export const enumFromStringExists = <T>({
```
mstrasinskis, line 1: Just thinking, will this add any value here?

```suggestion
export const enumFromStringExists = <T extends object>({
```

src/lib/utils/theme.utils.ts
```diff
@@ -0,0 +1,56 @@
+import { Theme } from "../types/theme";
+import { enumFromStringExists } from "./enum.utils";
+
+const isNode = (): boolean =>
```
mstrasinskis, line 4: Is there a plan to include `@dfinity/ic-js/utilis` into `gix-component`?

src/lib/utils/theme.utils.ts
```diff
@@ -0,0 +1,56 @@
...
+  // Jest NodeJS environment has no document
+  if (isNode()) {
+    return Theme.DARK;
+  }
+
+  const theme: string | null =
+    document.documentElement.getAttribute(THEME_ATTRIBUTE);
+
+  const initialTheme: Theme = enumFromStringExists({
+    obj: Theme as unknown as Theme,
```
mstrasinskis, line 22: I guess `as unknown as Theme` is not needed anymore. At least it doesn't complains locally...

src/tests/components/Toggle.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    expect(input.getAttribute("aria-label")).toEqual(props.ariaLabel);
+  });
+
+  it("should toggle checked", () => {
+    const { component, container } = render(Toggle, { props });
+
+    const input = container.querySelector("input") as HTMLInputElement;
+
+    const onToggle = jest.fn();
+    component.$on("nnsToggle", onToggle);
```
mstrasinskis, line 35: Thoughts aloud: probably it's time to switch to `gix` prefix for exposed events...

src/tests/utils/enum.utils.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { Theme } from "$lib/types/theme";
+import { enumFromStringExists } from "$lib/utils/enum.utils";
+
+describe("enum-utils", () => {
+  it("should return enum exists", () => {
+    expect(
+      enumFromStringExists<Theme>({
```
mstrasinskis, line 7: I guess `<typeof Theme>` would be a more precise type.

mstrasinskis (approved):
LGTM, thank you! Added a couple of minor optional comments.

src/lib/utils/enum.utils.ts
peterpeterparker: Not sure but I like it 😉

src/lib/utils/theme.utils.ts
peterpeterparker: I'm not sure, `utils` has a peer dependencies on `agent-js` which is not use here. Let's discuss it!

src/lib/utils/theme.utils.ts
peterpeterparker: Beautiful

src/tests/utils/enum.utils.spec.ts
peterpeterparker: Nice 👍

src/tests/components/Toggle.spec.ts
peterpeterparker: That's a cool idea. I just tried, renaming the events prefix will be easy. On the contrary, migrating the event listener would be quite some work because it is not typed

e.g. `<Back slot="back" on:nnsBack />`

There is no way with `npm run build` or `npm run check` to catch that `nnsBack` would have been renamed to `gixBack`. Only way to catch it automatically is if a particular test `npm run test` has been written.

So feels quite challenging. What do you think?

PR #82, opened 2022-10-10
style: dialog bigger height and wizard form

peterpeterparker:
# Motivation

The dialog modal new design is too wide for Transactions in NNS-dapp on larger (> medium) screen.

We can also deprecate the wizard style and generally style form in modal.

# Changes

- make dialog height fit-content on large screen
- set previous max-height as mix-height to smooth the opening transition
- replace wizard style with form style

# Screenshots

Is:

Capture d’écran 2022-10-10 à 07 40 44
Capture d’écran 2022-10-10 à 07 40 46

Will be:

Capture d’écran 2022-10-10 à 07 40 29
Capture d’écran 2022-10-10 à 07 40 35

### DIFF
--- a/src/lib/styles/global/modal.scss
+++ b/src/lib/styles/global/modal.scss
@@ -9,7 +9,7 @@ div.modal {
     display: flex;
     flex-direction: column;
     align-items: stretch;
-    justify-content: center;
+    justify-content: flex-start;
     gap: var(--padding-3x);
   }
 
--- a/src/lib/styles/global/modal.scss
+++ b/src/lib/styles/global/modal.scss
@@ -5,19 +5,12 @@
  */
 
 div.modal {
-  .wizard-wrapper {
+  form {
     display: flex;
     flex-direction: column;
     align-items: stretch;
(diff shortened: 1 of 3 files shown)


src/lib/styles/global/modal.scss
```diff
@@ -5,19 +5,12 @@
  */
 
 div.modal {
-  .wizard-wrapper {
+  form {
```
lmuntaner, line 8: Isn't this dangerous?

Any form inside the modal will behave the same way. We wanted it or not...

lmuntaner (approved):
Approving, but I added one concern.

src/lib/styles/global/modal.scss
peterpeterparker: I thought about it too - i.e. using a style class - but at the end of the day, we indeed probably want all form to look the same.

Not dangerous in any case because it only has to do with new design.

lmuntaner: Us yes, but not necessarily all the users of gix-components might want that.

PR #94, opened 2022-10-26
style: fix glitchy menu appearance on xtra large screen

peterpeterparker:
# Motivation

Do now wait on JavaScript to move the content once the menu is displayed but amend the fact that the menu will use space.


lmuntaner (approved):
Rubber stamp

PR #101, opened 2022-11-02
feat: new layout with a menu reveal and header positioning

peterpeterparker:
# Motivation

Implement the new layout design that replaces the overlay menu with a reveal type menu and make the header sticky only on smaller devices in a backwards compatible way - i.e. implement the new layout in existing design.

In addition, drop javascript usage to make the menu sticky to a css only approach (🥳).

# Screenshots

menuxlarge
menusmall


mstrasinskis (approved):
LGTM, thank you!

PR #103, opened 2022-11-05
style: fonts according new design

peterpeterparker:
# Motivation

Review and use the fonts according new design.

⚠️ The base font size goes down from browser default `16px` to `14px` as requested by the designer and therefore all fonts will appear smaller ⚠️

# Changes

- remove letter spacing for title
- remove ultra small size
- replace Circular bold by book
- replace "bold" font-weight 700 by 450
- downsize base font to 14px
- review and adapt all font-size and line-height h1, h2, etc.
- extract titles and small style to a mixins (can be useful when we want to style other elements in same style)

# Guidelines

Guidelines provided by the designer. 

image


peterpeterparker:
Instead of screenshots, let's have a chat / demo @mstrasinskis if you wish

mstrasinskis (approved):
LGTM, thank you!

PR #105, opened 2022-11-07
style: new design colors, spacing and layout

peterpeterparker:
# Motivation

Implement an extrapolated version of the new design provided by the designer without changes in the intrinsic behavior of the components.

# Changes

- review all theming CSS
- introduce a new `Content` component
- review `Tag` component
- few new assets (imported as such and resolved by Vite automatically - no conversion to `.svelte` cmp) and icons
- update jest config to load new svg and png assets

# Screenshots

No screenshots. Test version provided internally.


jest-transform.cjs
```diff
@@ -0,0 +1,12 @@
+const path = require("path");
```
lmuntaner, line 1: Why is this needed now? It wasn't needed before for svg assets right?

src/docs/styles/docs-global.scss
```diff
@@ -56,12 +56,21 @@ main {
...
     margin: var(--padding-3x) 0;
   }
+
+  table {
+    background: var(--card-background);
+  }
+
+  pre[class*="language-"] {
+    // We always want dark code block
+    background: #34294e;
```
lmuntaner, line 66: Can we use a variable here?

src/lib/components/Content.svelte
```diff
@@ -0,0 +1,85 @@
...
+
+  export let back = false;
+
+  // Enhance UI contrast for readability
+  export let contrast = false;
+</script>
+
+<div
+  class="content"
+  style={`--layout-bottom-offset: ${$layoutBottomOffset}px`}
```
lmuntaner, line 15: Why is this variable set inline?

lmuntaner (commented):
Just a few comments, leaving the "real" CR for @mstrasinskis  😄

jest-transform.cjs
peterpeterparker: Because this PR uses assets https://github.com/dfinity/gix-components/tree/style/new-design-colors-spacing-layout/src/lib/assets

Not svg encapsulated as svelte component but real assets which are `import ... from ....svg` with Vite and therefore are also resolved when the lib is packaged.

src/docs/styles/docs-global.scss
peterpeterparker: I did not find any passing existing variable in light mode and did not wanted to add a variable in the lib as this code block is only meant for the doc showcase - should not exists in the lib. That's why I used a fix value and added a comment.

src/lib/components/Content.svelte
peterpeterparker: Because the CSS style needs a dynamic binded value calculated with JavaScript. Not sure I get your question, how would you solve this differently?

Note: this is not new but refactored.

lmuntaner: I don't know, but this is the 8th way we have to change style.

https://www.notion.so/Reusing-styles-06430b86e1f14e698dd7b44a99d2b6c2

Maybe this is the only way to fix it.

I also think we should push back designing some things if they add a big layer of complexity.

peterpeterparker: Haha I still don't get your point here 😅. Not sure it is worth investing that deeper here, if you want to discuss it in a chat ping me anytime.

lmuntaner: No point for this PR, no worries. I am just not a fan of having so many ways of changing styles, and we added one more. Maybe it was the only way here. Just a comment to bring that topic, but I don't think you and @mstrasinskis are worried about that, so that's good.

peterpeterparker: We have developed this particular piece of code together with @mstrasinskis 😜

src/lib/components/Modal.svelte
```diff
@@ -27,7 +27,7 @@
 {#if visible}
   <div
     class="modal"
-    transition:fade
+    transition:fade={{ duration: 150 }}
```
mstrasinskis, line 30: Can any of 2 other values be applied here, to have the same ani.time everywhere, or it looks better with 150?
```
  --animation-time-normal: 0.2s;
  --animation-time-short: 0.125s;
  ```

File: src/lib/components/Back.svelte [hunk 5]
```diff
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import IconBackIosNew from "$lib/icons/IconBackIosNew.svelte";
+  import IconBack from "../icons/IconBack.svelte";
```
mstrasinskis, line 2: `$lib/`?

src/lib/components/Modal.svelte
peterpeterparker: Sure 125 works out

src/lib/icons/IconPassword.svelte
```diff
@@ -0,0 +1,15 @@
+<!-- source: DFINITY foundation -->
+<svg
+  width="20"
```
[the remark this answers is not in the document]
peterpeterparker: As discussed, designer choice

PR #107, opened 2022-11-07
build: auto organize import

peterpeterparker:
# Motivation

As for the dapps and libs, auto organize import with prettier (except in svelte component because not yet supported by the plugin).


mstrasinskis (approved):
Nice, thank you!

PR #109, opened 2022-11-09
style: banner offset

peterpeterparker:
# Motivation

On large screen the content needs also to amend the banner height (when displayed) in its offset.

# Screenshots

Is

Capture d’écran 2022-11-09 à 07 12 03

Will be

Capture d’écran 2022-11-09 à 07 11 27


lmuntaner (approved):
LGTM

PR #113, opened 2022-11-15
feat: backdrop over content if menu open on small devices

peterpeterparker:
# Motivation

- improve user experience when the menu pushes the content on mobile devices
- improve user experience glitch when menu is open on small window, then user navigate and browser is resized and back button is still displayed

# Changes

- use the `<Backdrop />` component in the `<SplitPan />` but animate with CSS only and not with JavaScript / Svelte to have a smooth behavior without the need to observe window size with JS

# Edge case

If user as a small window, open the menu, make the window bigger, then navigate, then make the window smaller, the "back" button and menu open are going to be displayed. I think it's an acceptable edge case - i.e. a pure CSS solution is better than a JavaScript less smooth solution.

# Screenshots

Enregistrement de l’écran 2022-11-15 à 10 46 47
Enregistrement de l’écran 2022-11-15 à 10 47 35

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@dfinity/gix-components",
-  "version": "2.0.0",
+  "version": "1.2.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@dfinity/gix-components",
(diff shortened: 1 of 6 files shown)


mstrasinskis (approved):
LGTM, thank you!

PR #116, opened 2022-11-16
style: dialog max height on iOS

peterpeterparker:
# Motivation

Fix max height for dialog on iOS.

# Screenshots

Before

Capture d’écran 2022-11-16 à 18 43 46

After

Capture d’écran 2022-11-16 à 18 43 50


lmuntaner (approved):
Thanks!

PR #118, opened 2022-11-17
fix: menu backdrop accessibility and z-index

peterpeterparker:
# Motivation

While displaying a backdrop over the content when the menu is open works, the accessibility is not good because the related element remains in the DOM even if not presented. This was an original decision for performance reason but after thinking about it, it's cleaner to remove it when not presented.

In addition, this PR also fixes a z-index issue.

# Changes

- add backdrop to dom over content only when needed
- close "menu open store" on wide screen with JS (to solve above requirements)
- fix z-index quirks

# Screenshots

image

### DIFF
--- a/src/lib/components/Content.svelte
+++ b/src/lib/components/Content.svelte
@@ -37,7 +37,9 @@
   </header>
 
   <div class="scrollable-content" class:open={$layoutMenuOpen}>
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
LGTM, thanks!

PR #120, opened 2022-11-17
style: summary grid

peterpeterparker:
# Motivation

For the new design we need some sort of "summary grid".

Might not be exactly what's in the design but, it's a start.

# Changes

- new media breakpoints for xlarge that amend the menu width
- new `.summary-grid`

# Screenshots

Design:

Capture d’écran 2022-11-17 à 17 14 35

Enregistrement de l’écran 2022-11-17 à 17 07 58


peterpeterparker:
On hold. Will probably never happen.

PR #124, opened 2022-11-18
style: badge color

peterpeterparker:
# Motivation

In new design badges have a custom linear gradient background.

Capture d’écran 2022-11-18 à 11 42 39


mstrasinskis (approved):
Rubber stamp

peterpeterparker:
Discussed with Mischa. We cancel this as displaying the badge with a background looks like a button which does not make sense in the context.

PR #130, opened 2022-11-23
style: better disable card color ratio

peterpeterparker:
# Motivation

As reported on [twitter](https://twitter.com/DfinityToday/status/1595439754396729344), the ratio when the card is disabled is not good enough.

image


nmattia (approved):
Rubber stamp

PR #134, opened 2022-11-24
feat: island

peterpeterparker:
# Motivation

Adds a new `<Island />` component that can be use to scope a container within the content.

In other words: "the scrollable content within the content that should be made not scrollable the designer has designed"

# Changes

- new `Component`, style and adjustment needed to integrate it in NNS-dapp

# Screenshots

Capture d’écran 2022-11-24 à 18 45 34
Capture d’écran 2022-11-24 à 18 45 37
km
Capture d’écran 2022-11-24 à 18 45 41
Capture d’écran 2022-11-24 à 18 45 44
Capture d’écran 2022-11-24 à 18 45 51
Capture d’écran 2022-11-24 à 18 45 56


src/lib/components/Island.svelte
```diff
@@ -0,0 +1,57 @@
+<script lang="ts">
+  import { onDestroy } from "svelte";
+  import { layoutContentScrollY } from "../stores/layout.store";
+
+  // 1024 is the $breakpoint-large size.
```
lmuntaner, line 5: Maybe create a variable for 1024?

```
const BREAKPOINT_L = 1024;
```

lmuntaner (approved):
Just a minor comment. LGTM!

src/lib/components/Island.svelte
peterpeterparker: Good idea 👍

PR #137, opened 2022-11-29
style: cell island and content grid

peterpeterparker:
# Motivation

Introduces a new style class to style cells as "Island" and review content grid cell spreading and gap as in new design.

# Note
In `Content` component the option `contrast` is not used anymore in NNS-dapp. However, just in case we would need it again soon, I did not removed it in this PR.

### DIFF
--- a/src/lib/components/Island.svelte
+++ b/src/lib/components/Island.svelte
@@ -20,15 +20,15 @@
 
 <style lang="scss">
   @use "../styles/mixins/media";
+  @use "../styles/mixins/content";
 
   .island {
-    background: var(--background);
-    color: var(--background-contrast);
+    @include content.island;
 
     height: 100%;
 
     ::-webkit-scrollbar {
-      background: var(--background);
+      @include content.island;
     }
 
     overflow: hidden;
--- a/src/lib/components/Island.svelte
+++ b/src/lib/components/Island.svelte
@@ -22,13 +22,13 @@
   @use "../styles/mixins/media";
 
   .island {
-    background: var(--focus-background);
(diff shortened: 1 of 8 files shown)


src/lib/components/Island.svelte
```diff
@@ -20,15 +20,15 @@
...
   .island {
-    background: var(--focus-background);
-    color: var(--focus-background-contrast);
+    @include content.island;
 
     height: 100%;
 
     ::-webkit-scrollbar {
-      background: var(--focus-background);
+      @include content.island;
```
lmuntaner, line 31: Why is this needed?

lmuntaner (approved):
Just a comment. already approving.

src/lib/components/Island.svelte
peterpeterparker: Because the component and the utility share the same colors. As we don't have component related style in the dark/light/variable scss, I used a mixin to avoid duplication so that if one day we change the island colors, it changes the color in both.

lmuntaner: But why is it also needed inside `::-webkit-scrollbar`? It's already above

lmuntaner: Does it style the scrollbar in the same way of the container and that's needed?

peterpeterparker: It styles the background of the scrollbar. The "color" field of the mixin is not use but including it spare duplicating the background property

Capture d’écran 2022-11-29 à 12 58 20

lmuntaner: ok, thanks for the clarification

PR #139, opened 2022-11-29
style: secondary button

peterpeterparker:
# Motivation

I noticed that the "color" of the secondary button was not the contrast one in the Figma file defined by the designer.

This PR tweaks the secondary button to be closer to what they have in mind.

### DIFF
--- a/src/lib/styles/mixins/_overlay.scss
+++ b/src/lib/styles/mixins/_overlay.scss
@@ -1,0 +1,5 @@
+@mixin colors {
+  background: var(--overlay-background);
+  color: var(--overlay-background-contrast);
+  --button-secondary-background: var(--focus-background);
+}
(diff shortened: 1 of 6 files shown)


src/lib/styles/mixins/_overlay.scss
```diff
@@ -0,0 +1,5 @@
+@mixin colors {
```
lmuntaner, line 1: why the name `overlay` for this mixin?

lmuntaner (approved):
LGTM, just a question

src/lib/styles/mixins/_overlay.scss
peterpeterparker: because it defines the background colors of the toast and modal

lmuntaner: aahh, ok, and they are overlays. Understood.

PR #141, opened 2022-12-28
Fix: Do not disconnect before update

lmuntaner:
# Motivation

InfiniteScroll didn't trigger intersection always.

# Changes

* Do not disconnect observer before update.

### DIFF
--- a/src/lib/components/InfiniteScroll.svelte
+++ b/src/lib/components/InfiniteScroll.svelte
@@ -43,9 +43,6 @@
     options
   );
 
-  // We disconnect previous observer before any update. We do want to trigger an intersection in case of layout shifting.
-  beforeUpdate(() => observer.disconnect());
-
   afterUpdate(() => {
     // The DOM has been updated. We reset the observer to the current last HTML element of the infinite list.
 


lmuntaner:

Was the reason to disconnect before update only to avoid a layout shift? I found a bug on mainnet that it's not requesting the voting history in the neuron details page.

I was able to replicate it by:
1. Refresh in a neuron detail page
2. Scroll down to the voting history
3. The voting history is updated correctly with infinite scroll
4. Go back to neurons page.
5. Click on the same neuron as before.
6. Scroll down to the voting history.
7. The voting history is not updated with the infinite scroll.

peterpeterparker:
As described in code

> We disconnect previous observer before any update. We do want to trigger an intersection in case of layout shifting.

i.e. we do not want to trigger unnecessary query and update on proposal list views.

lmuntaner:
The problem was that the observer somehow didn't connect again then...

I'll keep playing, but have you replicated the issue in mainnet?

peterpeterparker:
Not yet, I'll have a look tomorrow or next week

lmuntaner:
ok, no worries.

peterpeterparker:
@lmuntaner thanks for the reproduction details, very useful 👍

I found the root cause of the issue and implemented a workaround so that it preserves the existing feature which is useful for the "Proposal list" view 👉 https://github.com/dfinity/gix-components/pull/142

Therefore I close this PR but for sure reopen it if I am missing something.

lmuntaner:
@peterpeterparker thank you, that's perfect!

PR #151, opened 2023-01-19
feat: expose layout store

peterpeterparker:
# Motivation

Expose layout store so that we can use such information as the menu being open or not in NNS-dapp.


mstrasinskis (approved):
LGTM, thanks!

PR #154, opened 2023-02-08
feat: QR code and segment

peterpeterparker:
# Motivation

New QR Code and Segment components.

# Changes

- new `<QRCode />` component
- new `<Segment />` and `<SegmentButton />` components
- copy `debounce` from NNS-dapp
- context and store for segments
- docs and showcase accordingly
- introduces [Playwright](https://playwright.dev/) for e2e tests

# Tests

Screenshots tests for both segment and QR code have been implemented in this PR.
Automation with CI and Docker is not provided, we can add this in a following PR.

# Screenshots

See related implementatin and PR in NNS-dapp: https://github.com/dfinity/nns-dapp/pull/1842


package.json
```diff
@@ -47,7 +49,8 @@
   },
   "type": "module",
   "dependencies": {
-    "dompurify": "^2.4.3"
+    "dompurify": "^2.4.3",
+    "qr-creator": "^1.0.0"
```
lmuntaner, line 53: Does this need a security review?

src/lib/components/QRCode.svelte
```diff
@@ -0,0 +1,115 @@
...
+  });
+
+  const dispatch = createEventDispatcher();
+
+  const renderCanvas = () => {
+    if (canvas === undefined || size === undefined) {
+      return;
+    }
+
+    QrCreator.render(
```
lmuntaner, line 60: Stupid question: How does QrCreator know that it must leave space in the center for the logo?

src/lib/components/QRCode.svelte
```diff
@@ -0,0 +1,115 @@
...
+  import { debounce } from "$lib/utils/debounce.utils";
+
+  export let ariaLabel: string | undefined = undefined;
+  export let value: string;
+
+  // Valid CSS colors
+  export let fillColor = "black";
+  export let backgroundColor = "white";
+
+  export let radius = 0;
```
lmuntaner, line 19: What does `radius` do?

src/lib/components/Segment.svelte
```diff
@@ -0,0 +1,121 @@
...
+      width: currentClientWidth,
+    };
+  };
+
+  $: selectedElement, (() => initIndicator())();
+
+  let segmentsCount = 0;
+  $: segment,
+    (() =>
+      (segmentsCount =
```
lmuntaner, line 61: If a button is added dynamically, will it be taken into account?

`segment` variable will be the same HTMLElement; therefore, it won't change

It might not be necessary to solve it in this PR, but it's worth adding a TODO and solving afterwards?

src/lib/components/Segment.svelte
```diff
@@ -0,0 +1,121 @@
...
+    border-radius: var(--border-radius);
+    overflow: hidden;
+
+    width: var(--segment-min-width, var(--segment-width, fit-content));
+    --segment-min-width: calc(var(--segment-button-width) * var(--segments));
+
+    position: relative;
+  }
+
+  .indicator {
```
lmuntaner, line 103: If I understand correctly, this is the indicator of the current selected element?

Why didn't you add the style in the child instead? You have the element there, wouldn't it be easier?

lmuntaner (commented):
A few comments

package.json
peterpeterparker: Yes and no (they said it does not need particular review kind of) but in any case, security is aware of it. I asked them before even starting the process and informed them about what was implemented.

src/lib/components/QRCode.svelte
peterpeterparker: There is no space. The logo is opaque and above the QR code.

src/lib/components/QRCode.svelte
peterpeterparker: Added a comment.

> // The edge radius of each module. Must be between 0 and 0.5.

Did not implement particular rule. We can do that later if we need but currently we do not use it.

src/lib/components/Segment.svelte
peterpeterparker: Sure, added todo

src/lib/components/Segment.svelte
peterpeterparker: Because the animation is easier to implement with a single indicator `div` that moves along the all segment length (see GIF in https://github.com/dfinity/nns-dapp/pull/1842).

If we add the indicator in each segment button, then each shall moves on click which makes the animation logic a bit more tricky. Not against, if you want to implement it in another PR for sure that's cool.

package.json
lmuntaner: Ok, that's what matters.

src/lib/components/QRCode.svelte
lmuntaner: And that doesn't bother when you scan the QR code?

src/lib/components/Segment.svelte
lmuntaner: I understand. For the transition that makes sense.

lmuntaner (approved):

src/lib/components/QRCode.svelte
peterpeterparker: Yes and no. The QR code contains a repetition of the code (x times the same code) so technically, you can display a logo because the code is repeated but, if the bigger the logo the more difficult for the QR code reader to read the logo. I warned the designers about it.

lmuntaner: Ok, thanks for the explanation. I'm not an expert on QRs.

PR #155, opened 2023-02-12
build: make qr-creator available only in the browser

peterpeterparker:
# Motivation

The library leads to issues (es modules import error, segmentation fault, blocking tests etc.) in jest tests of NNS-dapp when use explicitly or imported implicitly.
Therefore, the simplest way to avoid these problems is to skip it globally in jest tests.
It remains tested in e2e tests.

P.S.: Mocking it would probably be possible but that way no one using the library as to care about this issue.

This PR solves both `npm run test` and `npm run build` of NNS-dapp.

# Changes

- lazy load the `qr-creator` library for the browser only


lmuntaner (approved):
Approving to unblock it.

But I guess we need a better solution? This will happen with any library we try to use, right?

peterpeterparker:
> This will happen with any library we try to use, right?

Most probably not. It's particular.

peterpeterparker:
Thanks for the review!

PR #158, opened 2023-02-18
fix: collapsible calculated height on Firefox

peterpeterparker:
# Motivation

A user noticed and reported on the [forum](https://forum.dfinity.org/t/modify-nns-front-code/18560) an issue with the `Collapsible` component which did not computed the expandable height correctly. I was able to reproduce it in Firefox only.

# Changes

- fix height computation of the `Collapsible` element by using primarly the `getBoundingClientRect` API which according test seems more accurate on Firefox and compatible with Chrome, Safari and iOS

# Screenshots

Bug

bug

Fix

fix

### DIFF
--- a/src/lib/components/Collapsible.svelte
+++ b/src/lib/components/Collapsible.svelte
@@ -17,7 +17,7 @@
   const dispatch = createEventDispatcher();
 
   export let expanded = initiallyExpanded;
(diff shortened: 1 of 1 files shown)


mstrasinskis (approved):
Interesting bug, thanks for fixing!

PR #161, opened 2023-03-01
feat: toast position top, title and custom icon

peterpeterparker:
# Motivation

Introduce few new options to display toasts "warning" in NNS-dapp. This includes a new position "end" - top-right on desktop and top-centered on mobile - an optional title, custom icon or an option to truncate text in case of overflow instead of scroll.

# Changes

- add new options to toast definition and `<Toast />` component
- add position option to `<Toasts />` so that two wrappers can be declared for and within an app
- set toast icon size (noticed the circle check was displayed bigger than other icon)
- new `IconMeter` 
- improve typescript declaration of toast store
- fix `id` usage in function `show` of toast store

# Screenshots


mstrasinskis (approved):
LGTM, thank you!

PR #162, opened 2023-03-01
feat: hide close button if modal disabled

peterpeterparker:
# Motivation

As requested, hide "Close" button if modal has disabled events (instead of disabling the button).


mstrasinskis (approved):
LGTM, thanks!

PR #164, opened 2023-03-03
feat: add an option to clean selected toasts

peterpeterparker:
# Motivation

We will use this option to clean selected toast after sign-in in NNS-dapp (insted of reset() and workaround).


lmuntaner (approved):
Thanks!

PR #172, opened 2023-03-16
style: button card

peterpeterparker:
# Motivation

New style for "button card".

# Note

PR also bump version of gix-cmp to release after merge.

# Changes

- new style `button.card`
- few variable refactored to become global, notably card border size
- grid height set to `1fr` to be consistent

# Screenshots

Capture d’écran 2023-03-16 à 08 32 48
Capture d’écran 2023-03-16 à 08 32 49
Capture d’écran 2023-03-16 à 08 32 54
Capture d’écran 2023-03-16 à 08 32 55

### DIFF
--- a/src/lib/styles/global/button.scss
+++ b/src/lib/styles/global/button.scss
@@ -78,12 +78,14 @@ button {
     }
   }
 
+  --primary-background: linear-gradient(
+    80.26deg,
+    var(--primary) 4.98%,
+    var(--secondary) 90.33%
+  );
+
   &.primary {
-    background: linear-gradient(
-      80.26deg,
-      var(--primary) 4.98%,
-      var(--secondary) 90.33%
-    );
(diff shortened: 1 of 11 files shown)


src/lib/styles/global/button.scss
```diff
@@ -132,4 +134,44 @@ button {
       filter: contrast(1.25);
     }
   }
+
+  &.card {
```
lmuntaner, line 138: Stupid question, but where's the background?

Shouldn't there be a place that sets background to `transparent` when not focus, nor active, no hover?

lmuntaner (approved):
I added a question, but so far lgtm and I'm approving. Thanks!

src/lib/styles/global/button.scss
peterpeterparker: Default = no background = transparent

PR #173, opened 2023-03-20
feat: mirror camera for QR code reader on desktop

peterpeterparker:
# Motivation

Mirror front facing camera on desktop.

# Notes

- MacOS does not provide any information about which camera was effectively used to create the stream (`facingMode = undefined`). That's why we fallback to an optimistic approach. If not mobile, then front camera assuming most desktop have only a front camera

- This PR only changes the display of the video for such use cases. The canvas and QR code are still decoded the same way. This is a visual change only.


dskloetd (approved):
Thanks! No test?

peterpeterparker:
I originally did not add an e2e test because it feels a bit like testing that a mock is a mock but, not against. So added an additional screenshot in the e2e suite.

PR #175, opened 2023-03-20
feat: trim QR code reader result

peterpeterparker:
# Motivation

This morning I noticed some leading spaces in a QR code I scanned. Probably worth trimming the result so that user will not have to edit the input field afterwards to remove such empty characters.


lmuntaner (approved):
Thanks!

PR #177, opened 2023-03-22
fix: QR code reader memory leak

peterpeterparker:
# Motivation

According our tests on mobile devices, it seems that `jsQR` library leaks memory on both Android and iOS.

Therefore this PR removes the implementation and the library and replace it with [html5-qrcode](https://github.com/mebjas/html5-qrcode).


lmuntaner (approved):
Wow, so much nicer! Thanks!

PR #178, opened 2023-03-28
Pass testId from WizardModal to Modal

dskloetd:
# Motivation

Currently, getting the root element of a WizardModal component by test ID requires adding an additional wrapping `div` to put the `data-tid` attribute on.
Since WizardModal already has a single root component, being able to set a test ID on that would remove the requirement of the extra div.

# Changes

1. Pass testId from WizardModal to Modal
2. Add data-tid test to Modal
3. Add data-tid test to WizardModal

# Screenshots

no visible changes

# Tests

Unit tests for Modal and WizardModal

### DIFF
--- a/src/lib/components/WizardModal.svelte
+++ b/src/lib/components/WizardModal.svelte
@@ -6,6 +6,7 @@
 
   export let steps: WizardSteps;
   export let disablePointerEvents = false;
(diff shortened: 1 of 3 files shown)


peterpeterparker (approved):
Cool, thanks 👍

PR #180, opened 2023-04-03
fix: menu background loading glitch

peterpeterparker:
# Motivation

Despite loaded in root, the actual default value of the theme store is undefined.

- when built where no DOM exists, therefore cannot guess the theme
- when loaded in the browser, because of the chunking, might be loaded after other resources have already been loaded and displayed

This is what lead the menu background to be displayed in dark mode over a light theme.

# Changes

- set `undefined` as default theme store value
- do not display menu background if theme is not set

# Note

I tried to add a `transition:fade` to makes the images appearing smooth but it had no effect.

# Screenshots

Issue:

issue

Solve:

solve

### DIFF
--- a/src/lib/components/MenuBackground.svelte
+++ b/src/lib/components/MenuBackground.svelte
@@ -35,7 +35,8 @@
       loading="lazy"
     />
 
-    <img data-tid="menu-background"
+    <img
+      data-tid="menu-background"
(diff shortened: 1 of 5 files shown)


mstrasinskis (approved):
LGTM, thanks!

PR #181, opened 2023-04-03
build: run Playwright e2e tests in CI

peterpeterparker:
# Motivation

Run and check e2e tests on PRs.

# Changes

- add `e2e` to `checks`
- add new `snapshots` GA to update snapshots on demand


.github/workflows/checks.yml
```diff
@@ -73,8 +73,36 @@ jobs:
...
+  e2e:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 60
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Cache and restore samples
+        uses: actions/cache@v2
+        id: samples
```
dskloetd, line 85: What does "samples" refer to?

.github/workflows/checks.yml
```diff
@@ -73,8 +73,36 @@ jobs:
...
+        with:
+          path: ./samples/
+          key: ${{ runner.os }}-samples-${{ hashFiles('**/scripts/download-samples.sh') }}
+          restore-keys: |
+            ${{ runner.os }}-samples-
+      - name: Install dependencies
+        run: npm ci
+      - name: Install Playwright Browsers
+        run: npx playwright install --with-deps chromium
```
dskloetd, line 94: It seems if you run with Google Chrome instead of Chromium, you don't need this step (which is quite slow).
See https://github.com/dfinity/nns-dapp/pull/2225/commits/ae153b1a1663605f14374b0dcd02bdc1d21a6cae

.github/workflows/snapshots.yml
```diff
@@ -0,0 +1,31 @@
...
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: npm ci
+      - name: Install Playwright Browsers
+        run: npx playwright install --with-deps chromium
+      - name: Run Playwright tests
+        run: npm run e2e:snapshots
+      - name: Commit Playwright updated snapshots
+        uses: EndBug/add-and-commit@v9
+        # We don't want to commit formatting changes to main
```
dskloetd, line 26: What does formatting have to do with it?

.github/workflows/snapshots.yml
```diff
@@ -0,0 +1,31 @@
...
+        if: ${{ github.ref != 'refs/heads/main' }}
```
dskloetd, line 27: Can't you configure the whole workflow to not run on main?
Or why is this `if` only here and not at the top?

e2e/qr-code-generator.e2e.ts-snapshots/QR-code-renders-https-nns-ic0-app-1-linux.png
dskloetd, line 1: Is it useful to have different screenshots which only differ in the contents of the QR code?

e2e/qr-code-reader.e2e.ts-snapshots/Read-QR-throws-error-because-user-block-camera-1-linux.png
dskloetd, line 1: This screenshot looks identical to the previous one?

dskloetd (commented):
I feel like I'm missing something but the number of screenshots seems excessive.
Are you using them to test functionality and not just rendering/style?

.github/workflows/snapshots.yml
peterpeterparker: Copy/paste, good catch. Thx

.github/workflows/snapshots.yml
peterpeterparker: I don't think you can configure `workflow_dispatch` to not be available on `main`.

That's why I added a guard at the start of this job.

```
      - name: Fail if branch is main
        if: ${{ github.ref == 'refs/heads/main' }}
        run: |
          echo "This workflow should not be triggered with workflow_dispatch on main"
          exit 1
```

But also added the additional `if` line you commented "just in case" and also because we use `EndBug/add-and-commit` in NNS-dapp and II always with a `if` - to follow the same pattern.

peterpeterparker: If you know how to configure `workflow_dispatch` to not be available on main, of course happy to hear it!

e2e/qr-code-generator.e2e.ts-snapshots/QR-code-renders-https-nns-ic0-app-1-linux.png
peterpeterparker: Yes, two different goals. One test that a QR code is rendered, one test that changing the QR code value on the fly re-render the QR code.

Not related to this PR.

peterpeterparker: Screenshots were generated here because CI use "linux" while my Mac locally use "darwin"

e2e/qr-code-reader.e2e.ts-snapshots/Read-QR-throws-error-because-user-block-camera-1-linux.png
peterpeterparker: Almost, the font renderering is not exactly the same on linux and darwin therefore `Read-QR-throws-error-because-user-block-camera-1-linux.png` and `Read-QR-throws-error-because-user-block-camera-1-darwin.png` are pixel wise different.

.github/workflows/checks.yml
peterpeterparker: Cool, thanks. Done.

Notes: I regenerated all screenshots, darwin and linux

.github/workflows/checks.yml
dskloetd: Sorry for my confusion but what's in the samples folder? What's being cached?

e2e/qr-code-reader.e2e.ts-snapshots/Read-QR-throws-error-because-user-block-camera-1-Google-Chrome-linux.png
dskloetd, line 1: Is there any difference between 
`e2e/qr-code-reader.e2e.ts-snapshots/Read-QR-throws-error-because-camera-stream-fails-1-Google-Chrome-linux.png`
and
`e2e/qr-code-reader.e2e.ts-snapshots/Read-QR-throws-error-because-user-block-camera-1-Google-Chrome-linux.png`
?

dskloetd (approved):

e2e/qr-code-reader.e2e.ts-snapshots/Read-QR-throws-error-because-user-block-camera-1-Google-Chrome-linux.png
dskloetd: https://playwright.dev/docs/test-snapshots

> Alternatively you can specify snapshot name as the first argument of the `toHaveScreenshot()` method:

peterpeterparker: Nice ! Let's do that afterwards, thanks

.github/workflows/checks.yml
peterpeterparker: Answered https://github.com/dfinity/gix-components/pull/181#discussion_r1156127139

e2e/qr-code-reader.e2e.ts-snapshots/Read-QR-throws-error-because-user-block-camera-1-Google-Chrome-linux.png
peterpeterparker: https://github.com/dfinity/gix-components/pull/183

PR #183, opened 2023-04-04
test: spare one screenshot

peterpeterparker:
# Motivation

Spare the one screenshot that is similar for two e2e tests.

Follow-up https://github.com/dfinity/gix-components/pull/181#discussion_r1156146248.


e2e/qr-code-reader.e2e.ts
```diff
@@ -25,7 +27,9 @@ const testQRCode = async ({
   // Wait for video
   await page.waitForTimeout(2000);
 
-  await expect(page).toHaveScreenshot();
+  screenshotName !== undefined && screenshotName !== ""
```
dskloetd, line 30: Would the string ever be empty?

e2e/qr-code-reader.e2e.ts
```diff
@@ -25,7 +27,9 @@ const testQRCode = async ({
...
+    ? await expect(page).toHaveScreenshot(screenshotName)
+    : await expect(page).toHaveScreenshot();
```
dskloetd, line 32: Isn't passing undefined the same as passing nothing?

e2e/qr-code-reader.e2e.ts
```diff
@@ -55,7 +62,10 @@ test("Read QR throws error because camera stream fails", async () => {
 
   const page = await context.newPage();
 
-  await testQRCode({ page });
+  await testQRCode({
+    page,
+    screenshotName: "Read-QR-throws-error.png",
```
dskloetd, line 67: Is the screenshot there just to verify that the toast appeared? Maybe just check that the toast is there directly rather than through a screenshot?

e2e/qr-code-reader.e2e.ts
peterpeterparker: No but we cannot enforce it TypeScript wise.

e2e/qr-code-reader.e2e.ts
peterpeterparker: I wish too but unfortunately no, that's why I add to add this

e2e/qr-code-reader.e2e.ts
peterpeterparker: The test is there to test that an error was throw. We cannot test the QR code in a NodeJS context because the library requires a rendering window.

e2e/qr-code-reader.e2e.ts
dskloetd: Does TypeScript complain if you remove the check for `""`?

e2e/qr-code-reader.e2e.ts
dskloetd: Can you test that the toast is in the DOM and contains an error message?

e2e/qr-code-reader.e2e.ts
peterpeterparker: I mean Playwright fails if we path `""` as screenshot name.

dskloetd: But why would someone pass `""` as a screenshot name?
Aren't there lots of things you could pass that it would fail on?
Do you want to check all of them?
Like `"/"` or `"."` probably also fail, right?

dskloetd (approved):

e2e/qr-code-reader.e2e.ts
peterpeterparker: Removed the empty check

PR #196, opened 2023-04-13
style: modal form gap narrower

peterpeterparker:
# Motivation

The modal being narrower on desktop and given the fact that we have to display more information in the transactions modal, setting a smaller gap between lines of `form` within modal fits better the display.


mstrasinskis (approved):
Thank you!

PR #199, opened 2023-04-18
Feat: TestId changes in Collapsible and KeyValuePairInfo

lmuntaner:
# Motivation

To improve testing, the testId must be in the most parent element.

In KeyValuePairInfo that means that the Collapsible should have the test id.

# Changes

* Add prop testId in Collapsible.
* KeyValuePairInfo passes testId to parent Collapsible.

# Screenshots

* No new tests required.

### DIFF
--- a/src/lib/components/Collapsible.svelte
+++ b/src/lib/components/Collapsible.svelte
@@ -2,11 +2,12 @@
   import { afterUpdate, createEventDispatcher } from "svelte";
   import IconExpandMore from "$lib/icons/IconExpandMore.svelte";
   import { i18n } from "$lib/stores/i18n";
+  import TestIdWrapper from "./TestIdWrapper.svelte";
 
   export let id: string | undefined = undefined;
   export let initiallyExpanded = false;
   export let maxContentHeight: number | undefined = undefined;
-  export let testId: string = "collapsible-header";
+  export let testId: string | undefined = undefined;
 
   export let iconSize: "small" | "medium" = "small";
   export let expandButton = true;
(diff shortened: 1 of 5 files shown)


lmuntaner:
@peterpeterparker @dskloetd please review

src/lib/components/Collapsible.svelte
```diff
@@ -59,7 +60,7 @@
 </script>
 
 <div
-  data-tid="collapsible-header"
+  data-tid={testId ?? "collapsible-header"}
```
peterpeterparker, line 63: can it be simplified by declaring a default value?

```
export let testId = "collapsible-header";
```

peterpeterparker (approved):
LGTM, minor comment

src/lib/components/Collapsible.svelte
lmuntaner: oh, yes, thanks!

src/lib/components/Collapsible.svelte
```diff
@@ -59,7 +60,7 @@
 </script>
 
 <div
-  data-tid="collapsible-header"
```
dskloetd, line 62: This data-tid seems to be specific for the *header* as opposed to the data-tid="collapsible-content" on line 87. I think we should probably keep it and have a separate data-tid for the entire component.

src/lib/components/Collapsible.svelte
```diff
@@ -59,7 +60,7 @@
 </script>
 
 <div
```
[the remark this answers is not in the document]
lmuntaner: It's a good point. Yes, it might be something useful. I agree we shuold wrap it all with a div.

What I'm not so sure is whether this is a concern for gix-components or instead it should be done at the nns-dapp level.

We never discussed this, but coupling gix-components with nns-dapp is not the point of this library. I'm not even sure test ids should be here at all.

Instead, we might want to wrap those components in the nns-dapp level adding a test id.

src/lib/components/Collapsible.svelte
lmuntaner: done

src/lib/components/Collapsible.svelte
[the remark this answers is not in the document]
lmuntaner: done

lmuntaner:
@dskloetd please review

lmuntaner:
FYI: I tried the component with nns-dapp and collapsible looks the same in different use cases.

src/lib/components/Collapsible.svelte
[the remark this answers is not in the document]
dskloetd: Those are good questions that I cannot answer. If you prefer to do it on the nns-dapp side, that's fine with me but how would that be possible with gix components which are nested inside other gix components?

dskloetd (approved):

PR #207, opened 2023-04-26
fix: Warn: A11y: visible, non-interactive elements with an on:click event must be accompanied by an on:keydown, on:keyup, or on:keypress event.

peterpeterparker:
# Motivation

Fix warning:

> Warn: A11y: visible, non-interactive elements with an on:click event must be accompanied by an on:keydown, on:keyup, or on:keypress event. (svelte)

# Note about a11y

I just catched `Enter`, don't know if we should also catch `space bar` @mstrasinskis ?

# Changes

- `div` that behaves have button should:
   - `role=button`
   - `tabindex=0` or `-1`
   - `on:keypress`


src/lib/components/Collapsible.svelte
```diff
@@ -129,6 +130,12 @@
       align-items: center;
       justify-content: flex-start;
     }
+
+    outline: none;
+
+    &:focus {
+      filter: contrast(1.25);
```
mstrasinskis, line 137: Out of curiosity, is it noticeable on text?

src/lib/utils/keyboard.utils.ts
```diff
@@ -0,0 +1,13 @@
+export const onEnter = ({
+  $event: { key },
+  callback,
+}: {
+  $event: KeyboardEvent;
+  callback: () => void;
+}) => {
+  if ("Enter" !== key) {
```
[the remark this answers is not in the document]
peterpeterparker: Following?

```
export const handleKeyPress = ({
  $event: { code },
  callback,
}: {
  $event: KeyboardEvent;
  callback: () => void;
}) => {
  if (!["Enter", "Space"].includes(code)) {
    return;
  }

  callback();
};
```

src/lib/components/Collapsible.svelte
peterpeterparker: Not really but, did not had good idea. `text-decoration: underline;` or better idea?

mstrasinskis: Thanks, I think it's good as it is 👍

Maybe something for later to make tab-navigation focus more prominent e.g.
```
&:focus-visible {
  outline: var(--padding-0_25x) solid var(--warning-emphasis);
  outline-offset: calc(-1*var(--padding-0_25x));
}
```
image

src/lib/utils/keyboard.utils.ts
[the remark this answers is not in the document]
mstrasinskis: Looks good!

src/lib/components/Collapsible.svelte
peterpeterparker: That's a great idea! If you want to give it a try later, go for it

PR #210, opened 2023-05-09
style: number utility class

peterpeterparker:
# Motivation

Add a number utility class to display numbers monospaced.

### DIFF
--- a/src/lib/styles/global/utility-classes.scss
+++ b/src/lib/styles/global/utility-classes.scss
@@ -51,3 +51,9 @@
 .no-margin {
   margin: 0 !important;
 }
+
+// Format numbers as monospaced figures
+
+.num {
+  font-feature-settings: "tnum";
+}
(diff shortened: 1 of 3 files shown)


src/lib/styles/global/utility-classes.scss
```diff
@@ -51,3 +51,9 @@
 .no-margin {
   margin: 0 !important;
 }
+
+// Format numbers as monospaced figures
+
+.num {
```
mstrasinskis, line 57: nitpick: Should all the numbers be monospaced? If not, probably more precise name would fit better (`tabular-num`?)

mstrasinskis (approved):
LGTM, thanks!

src/lib/styles/global/utility-classes.scss
peterpeterparker: Makes sense sure, done

PR #212, opened 2023-05-12
fix: optionally format raw input to normal number format

mstrasinskis:
# Motivation

Avoid exponential formatting for the icp values.

Original issuer: when a user clicks on the 'Max' button and has a very small amount of tokens left, the value appears in exponential format, which is incorrect. The fix checks and replaces any Xe-Y formatted value on change.

# Changes

- apply optional formatting to input

# Tests

- should avoid exponent formatted on change in icp mode

# Screenshot

## Before
image

## After
Screen Recording 2023-05-11 at 10 27 07

### DIFF
--- a/src/lib/components/Input.svelte
+++ b/src/lib/components/Input.svelte
@@ -96,7 +96,7 @@
 
   const handleInput = ({ currentTarget }: InputEventHandler) => {
     if (inputType === "icp") {
-      const currentValue = currentTarget.value;
+      const currentValue = exponentToPlainNumberString(currentTarget.value);
 
       // handle invalid input
       if (isValidICPFormat(currentValue) === false) {
--- a/src/tests/lib/components/Input.spec.ts
+++ b/src/tests/lib/components/Input.spec.ts
@@ -396,6 +396,25 @@ describe("Input", () => {
       expect(container.querySelector("input")?.value).toBe("0.00000001");
     });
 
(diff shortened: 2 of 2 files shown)


src/tests/lib/components/Input.spec.ts
```diff
@@ -396,6 +396,25 @@ describe("Input", () => {
       expect(container.querySelector("input")?.value).toBe("0.00000001");
     });
 
+    it.only("should avoid exponent formatted on change in icp mode", async () => {
```
peterpeterparker, line 399: `.only` leftover?

src/lib/components/Input.svelte
```diff
@@ -96,7 +96,7 @@
 
   const handleInput = ({ currentTarget }: InputEventHandler) => {
     if (inputType === "icp") {
-      const currentValue = currentTarget.value;
+      const currentValue = exponentToPlainNumberString(currentTarget.value);
```
peterpeterparker, line 99: ah nice!

peterpeterparker (approved):
Already approving, checkout the comment about the test. Thanks!

src/tests/lib/components/Input.spec.ts
mstrasinskis: 🤦‍♂️ thank you!

dskloetd:
This PR seems to have broken the test:
https://github.com/dfinity/gix-components/actions/runs/4958375748/jobs/8871160346

PR #214, opened 2023-05-14
fix: input pre-formatting test

mstrasinskis:
# Motivation

One of the tests was broken due to non-synchronous input value formatting.

# Changes

- waitFor was added to wait for the right value

### DIFF
--- a/src/tests/lib/components/Input.spec.ts
+++ b/src/tests/lib/components/Input.spec.ts
@@ -4,7 +4,7 @@
 
 import Input from "$lib/components/Input.svelte";
 import { isNullish, nonNullish } from "@dfinity/utils";
-import { fireEvent, render } from "@testing-library/svelte";
+import { fireEvent, render, waitFor } from "@testing-library/svelte";
 import InputTest from "./InputTest.svelte";
 import InputValueTest from "./InputValueTest.svelte";
 
@@ -412,7 +412,7 @@ describe("Input", () => {
       }
 
       fireEvent.input(input, { target: { value: 0.00000001 } });
-      expect(input.value).toBe("0.00000001");
+      await waitFor(() => expect(input.value).toBe("0.00000001"));
     });
 
     it("should avoid exponent formatted initial value (>0) in icp mode", () => {


src/tests/lib/components/Input.spec.ts
```diff
@@ -412,7 +412,7 @@ describe("Input", () => {
       }
 
       fireEvent.input(input, { target: { value: 0.00000001 } });
```
dskloetd, line 414: Aren't input events always with strings?
Does clicking the "Max" button generate an input event? I would think it sets the value directly?

src/tests/lib/components/Input.spec.ts
```diff
@@ -412,7 +412,7 @@ describe("Input", () => {
...
-      expect(input.value).toBe("0.00000001");
+      await waitFor(() => expect(input.value).toBe("0.00000001"));
```
dskloetd, line 415: Why is the value `"1e-8"` before waiting (as can be seen from the failing test)?
I would think it should go straight from `""` to `"0.00000001"`.

I'm also curious how this got merged while this was the exact test being added in the PR that broke it.
Was there a conflict with another PR being merged at the same time?
Or some other kind of pitfall we could learn from?

src/tests/lib/components/Input.spec.ts
mstrasinskis: True, good point! Replaced with `$set`.

src/tests/lib/components/Input.spec.ts
mstrasinskis: > Why is the value "1e-8" before waiting (as can be seen from the failing test)?
I would think it should go straight from "" to "0.00000001".

### Description 
- The native element uses  `icpValue` instead of `value` -> `<input value={inputType === "icp" ? icpValue : value}...` 
- Expected that the type number of `value` is preserved.

**Firs commit logic:**
- Here where it was set to "1e-8" --> [icpValue = fixUndefinedValue(currentValue)](https://github.com/dfinity/gix-components/blob/c95a8168682bc1a3c346bafa8774f34b00482c21/src/lib/components/Input.svelte#L115);
- But on the next line it updates the value [value = +currentValue;](https://github.com/dfinity/gix-components/blob/c95a8168682bc1a3c346bafa8774f34b00482c21/src/lib/components/Input.svelte#L120)
- Then the test checks for the first time (FAIL)
- And then the block `$: value, ...` fixes the `icpValue` (which will be rendered).

I've played around with different solutions and apparently tested not with the committed one. The logic was fixed w/ the second commit.

**Second commit:**
- Fixes directly on the value update `icpValue = exponentToPlainNumberString(`${value}`);`

> I'm also curious how this got merged while this was the exact test being added in the PR that broke it.

It's weird, I can not reproduce a successful result w/ the original PR.

dskloetd: Thanks for explaining. Would it make sense to call `exponentToPlainNumberString` already immediately when we call `fixUndefinedValue` instead of allowing the value to become `"1e-8"` first and then fixing it asynchronously in the next round?

mstrasinskis: Unfortunately both functions can not be combined easily because of this block: https://github.com/dfinity/gix-components/blob/c95a8168682bc1a3c346bafa8774f34b00482c21/src/lib/components/Input.svelte#L104
Probably it will work, but the impact needs to be tested a lot. It would be better to simplify the logic and provide a type string only interface, but definitely not in this PR.

dskloetd (approved):

PR #215, opened 2023-05-15
Fix BottomSheet race condition

dskloetd:
# Motivation

`<BottomSheet>` appears inline on large screens and docked to the bottom on small screens.
When it's docked to the bottom, it sets `$layoutBottomOffset` so layout components can take it into account.
`BottomSheet` listens to `innerWidth` of the window and `clientHeight` of the component.
In practice `clientHeight` is set first and since `innerWidth` is initialized to `0`, the code acts as if the screen is small until `innerWidth` is also set.
This means that on large screens, `$layoutBottomOffset` is briefly set to non-zero and then immediately back to `0` when `innerWidth` is set.
On CI (although I wasn't able to reproduce this on my Mac) this consistently results in the page being rendered twice.
And for some reason the second render happens before the first page is destroyed.
On the swap participation page of NNS dapp, this means that `ProjectDetail` will start polling for the ticket twice (which is redundant) and then immediately stops polling for the ticket when `onDestroy` is called.
This results in a UI that's "loading forever".

# Changes

1. In `BottomSheet.svelte` initialize `height` and `innerWidth` to `undefined` instead of `0`.
2. Only update `$layoutBottomOffset` if both `height` and `innerWidth` are defiend.

# Tests

I tried to add a unit test for this but the jsdom unit test environment does not behave in a way that allows exposing the bug.

I did make a fork of `BottomSheet.svelte` in the nns-dapp repo to check that fixing the problem actually allows the sns participation end-to-end test to participate:
https://github.com/dfinity/nns-dapp/actions/runs/4967150795

### DIFF
--- a/src/lib/components/BottomSheet.svelte
+++ b/src/lib/components/BottomSheet.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import { nonNullish } from "@dfinity/utils";
   import { layoutBottomOffset } from "$lib/stores/layout.store";
   import { onDestroy } from "svelte";
   import { BREAKPOINT_LARGE } from "../constants/constants";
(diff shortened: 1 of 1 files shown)


dskloetd:
The broken test is because of https://github.com/dfinity/gix-components/pull/212

lmuntaner (approved):
Wow that was hiddne, thank you!!!

PR #218, opened 2023-05-26
Add testId to Value component

dskloetd:
# Motivation

I want to refer to a Value component from a page object in the nns-dapp repository.

# Changes

Add `testId` prop which sets `data-tid` attribute to `Value` component.

# Screenshots

N/A


mstrasinskis (approved):
LGTM

PR #220, opened 2023-05-31
fix: duplicate form field id in the same form

peterpeterparker:
# Motivation

Fix Chrome warning "uplicate form field id in the same form"

# Changes

- add a naive counter that provide an incremented id each time it is used

# Screenshot

Issue reported by Chrome.

Capture d’écran 2023-05-31 à 14 18 49


mstrasinskis (approved):
LGTM, nice solution!

PR #222, opened 2023-06-05
Modal data-tid on title

dskloetd:
# Motivation

I want to migrate `frontend/src/tests/lib/modals/neurons/MergeNeuronsModal.spec.ts` in nns-dapp to use page objects and I need a way to get to the title of a Modal for that.

# Changes

Add `data-tid="modal-title"` on the title slot in `Modal.svelte`.

# Tested

I ran `npm run dev` and verified that the title of the `WizardModal` component has the `datat-tid="modal-title"` attribute.

### DIFF
--- a/src/lib/components/Modal.svelte
+++ b/src/lib/components/Modal.svelte
@@ -51,7 +51,9 @@
     >
       {#if showHeader}
         <div class="header">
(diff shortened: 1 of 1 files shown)


mstrasinskis (approved):
LGTM, thanks!

PR #224, opened 2023-06-26
feat: icon sync & header buttons style

peterpeterparker:
# Motivation

Add an animated icon "Sync" and set a fixed with for the buttons in the header (that way it looks round on active)

# Notes

Screenshots are updated because there is a small bit more of margin at the start and end in the header.


lmuntaner (approved):
Thanks

PR #231, opened 2023-07-06
Feat: Page header

lmuntaner:
# Motivation

New component for the new neuron setting page: PageHeader

# Changes

* New component `PageHeader`.
* Add new component to docs.

# Screenshots

Screenshot of the doc page:

page-header-2


lmuntaner:

Still pending confirmation from design. But let me know what you think.

mstrasinskis (commented):
LGTM, but it would be great if @peterpeterparker looked too.

peterpeterparker:
> Still pending confirmation from design. 

Happy to review it when design is finalized.

mstrasinskis (approved):
Approved, as this pr is not directly related to the redesign.

lmuntaner:
> > Still pending confirmation from design.
> 
> Happy to review it when design is finalized.

I removed the comment and put it ready for review. But I should have made it clearer that design approved it.

@mstrasinskis approved it, but feel free to take a look

src/lib/assets/icp-rounded.svg
peterpeterparker, line 1: Why does this icon finds place in the lib?

src/lib/components/PageHeader.svelte
```diff
@@ -0,0 +1,43 @@
+<script lang="ts">
+  export let testId: string | undefined = undefined;
+
+  let hasHeader: boolean;
+  $: hasHeader =
+    $$slots.topLeft !== undefined || $$slots.topRight !== undefined;
```
peterpeterparker, line 6: We are generally using `start` and `end` as slot naming convention. 
Inherited from other design system that use such words when they support both RTL and LTR.

src/lib/components/PageHeader.svelte
```diff
@@ -0,0 +1,43 @@
...
+<style lang="scss">
+  .container {
+    display: flex;
+    flex-direction: column;
+    gap: var(--padding-4x);
+  }
+
+  .header {
+    display: flex;
+    flex-direction: row;
```
peterpeterparker, line 31: Isn't `row` the default?

src/routes/(split)/components/page-header/+page.md
```diff
@@ -0,0 +1,137 @@
...
+
+| Property | Description                                                     | Type                    | Default     |
+| -------- | --------------------------------------------------------------- | ----------------------- | ----------- |
+| `testId` | Add a `data-tid` attribute to the DOM, useful for test purpose. | `string` or `undefined` | `undefined` |
+
+## Slots
+
+| Slot name  | Description                                        |
+| ---------- | -------------------------------------------------- |
+| `title`    | The main information to render.                    |
```
peterpeterparker, line 30: I know I'm being picky here so sorry in advance for the question but, if it's an "information", why not call the slot "information"? 

Maybe use the word "title" for the description of the slot as well. Same for subtitle

src/lib/components/PageHeader.svelte
```diff
@@ -0,0 +1,43 @@
...
+<div class="container" data-tid={testId}>
+  {#if hasHeader}
+    <div class="header">
+      <slot name="topLeft" />
+      <slot name="topRight" />
+    </div>
+  {/if}
+  <div class="content">
+    <slot name="title" />
+    <slot name="subTitle" />
```
peterpeterparker, line 18: Is "subtitle" a word? should the slot name be lower case then?

src/routes/(split)/components/page-header/+page.md
```diff
@@ -0,0 +1,137 @@
...
+    <h1 slot="title">Title</h1>
+    <h3 slot="subTitle">This page header is inside a card</h3>
+  </PageHeader>
+</div>
+
+### Real production example
+
+<div class="wrapper">
+  <PageHeader>
+    <span slot="topLeft">
```
peterpeterparker, line 56: should it rather be a `svelte:fragment`?

peterpeterparker (changes requested):
Few comments. Just marking as it as "Requested changes" because already accepted.

src/lib/components/PageHeader.svelte
lmuntaner: makes sense, I'll rename

src/lib/components/PageHeader.svelte
lmuntaner: yes, I wanted to be specific.

src/routes/(split)/components/page-header/+page.md
lmuntaner: I was looking for synonyms of the name of the prop. Let me rephrase it.

src/lib/components/PageHeader.svelte
lmuntaner: yes, it is a word. Thanks

src/routes/(split)/components/page-header/+page.md
lmuntaner: This is an example, I used a `span`. It depends on who uses it.

src/lib/assets/icp-rounded.svg
lmuntaner: Where should I put it?

I will reuse it in other examples.

src/lib/components/PageHeader.svelte
peterpeterparker: It might be incorrect but I've got the feeling that the bundler does not clean the CSS of its default tags when build. Therefore I personally think it's best to not be "specific" with CSS as it makes the CSS heavier.

But maybe there is another valid reason? Happy to discuss it.

src/lib/components/PageHeader.svelte
lmuntaner: done

lmuntaner:
@peterpeterparker ready for next review

src/routes/(split)/components/page-header/+page.md
peterpeterparker: Kind of feel it would be cleaner to provide example with fragments.

The component has a structure and is opiniated, so the default should be to use it as it without using sub element when not needed.

Not a strong opinion though. Thanks for the answer.

src/lib/assets/icp-rounded.svg
peterpeterparker: - is it part of the lib and provide for those who use the lib? => it should be a svelte icon component
- is it use only in the document? => it should be move to `src/docs/assets` (`assets` folder to be created)

src/lib/components/PageHeader.svelte
lmuntaner: No strong opinion.

src/lib/assets/icp-rounded.svg
lmuntaner: the second. I'll move it

src/lib/components/PageHeader.svelte
peterpeterparker: then let's remove it if you don't mind

lmuntaner: done

src/lib/assets/icp-rounded.svg
lmuntaner: done

lmuntaner:
@peterpeterparker ready for another review

peterpeterparker (approved):
LGTM, thanks

Btw. if you want to be fancy and have fun in another PR, the library supports screenshot comparison testing. Not a must for this function, just saying really stricly if you want to have fun.

lmuntaner:
> LGTM, thanks
> 
> Btw. if you want to be fancy and have fun in another PR, the library supports screenshot comparison testing. Not a must for this function, just saying really stricly if you want to have fun.

How do I do that? I'll do it in another PR.

peterpeterparker:
Basically you just navigate to the page and take a screenshot.

https://github.com/dfinity/gix-components/blob/19e457542de039d4d648c40a59757aed164bf3f4/e2e/segment.e2e.ts#L10

You can generate the screenshot that is use for local testing and development with `npm run e2e:snapshots`.
To generate the screenshot once your PR exists, you can use the manual github trigger https://github.com/dfinity/gix-components/actions/workflows/snapshots.yml by just selecting your branch

I thought I already had documented this in HACKING.md but turns out not, so a good reminded I should still do it, thanks!

peterpeterparker:
Actually, I can document it and then when you do the review, you try to write a test. That way you bullet proof the documentation. I'll ping you.

lmuntaner:
> Actually, I can document it and then when you do the review, you try to write a test. That way you bullet proof the documentation. I'll ping you.

Sounds like a plan!

PR #232, opened 2023-07-07
Feat: Section Component

lmuntaner:
# Motivation

New component for the new neuron setting page: PageSection

**PENDING: Design approval**

# Changes

* New component PageSection.
* Add new component to docs.

# Screenshots

Screenshot of the documentation page with the showcase:

section-3

Screenshot of the mobile version of the real production showcase

Screenshot 2023-07-11 at 07 52 36


lmuntaner:
@mstrasinskis please review

src/lib/components/PageSection.svelte
```diff
@@ -0,0 +1,35 @@
+<script lang="ts">
+  export let title: string;
+  export let testId: string | undefined = undefined;
+</script>
+
+<div class="container" data-tid={testId}>
+  <div class="title-wrapper">
+    <h3 class="title">{title}</h3>
+    <slot name="value" />
```
mstrasinskis, line 9: Probably the slot `value` could have a more generic name. Or the component name could be more specific 🤔

src/lib/components/PageSection.svelte
```diff
@@ -0,0 +1,35 @@
+<script lang="ts">
+  export let title: string;
+  export let testId: string | undefined = undefined;
+</script>
+
+<div class="container" data-tid={testId}>
+  <div class="title-wrapper">
+    <h3 class="title">{title}</h3>
```
mstrasinskis, line 8: Class `.title` is not in use.

mstrasinskis: Would it make sense to provide a slot to be more flexible (h2, h4)?

mstrasinskis (commented):
Just a couple of comments. It feels a bit like some hierarchy information is missing. What is a page, Where can it be used. Which components are allowed to be on the page. Alternatively this could be a VerticalContainer with some slots...

src/lib/components/PageSection.svelte
lmuntaner: Yes, I renamed it to `end` following the convention that @peterpeterparker mentioned

src/lib/components/PageSection.svelte
lmuntaner: I wanted to discuss this with Artem. But it makes sense actually.

This way there can be sections of different levels maybe. I'll change it, thanks!

src/lib/components/PageSection.svelte
mstrasinskis: Thanks 👍

lmuntaner:
@mstrasinskis ready for another review.

I renamed it to Section

artkorotkikh-dfinity:
@lmuntaner can we please adjust a few spacings in this component to something like this? (I also adjusted on Figma): 
image

Questions: 
1. Is it safe to assume that this component will support things like "copy address" in `<span slot="value" class="value">`? 
image

lmuntaner:
@artkorotkikh-dfinity thanks for the review!

I did the changes.

Yes, the copy button can be added. That's another component unrelated to this one: `KeyValuePair`. It's already in this library and has worked bery well so far.

PR #237, opened 2023-07-10
Feat: ItemAction Component

lmuntaner:
# Motivation

Add a new component to be used in the new neuron settings page: ListItemAction.

PENDING: Design approval

# Changes

* New Component ItemAction.
* Add `size` prop to IconLockOpen.

# Screenshots

item-action-2

Mobile version:

Screenshot 2023-07-11 at 08 17 05


lmuntaner:

peterpeterparker:
> PENDING: Design approval

Still pending?

lmuntaner:
> > PENDING: Design approval
> 
> Still pending?

Yes, I shared it with Artem already though. We already discussed it, so I don't expect much changes.

src/lib/components/ListItemAction.svelte
```diff
@@ -0,0 +1,55 @@
+<div class="container">
```
peterpeterparker, line 1: If this component represent list and items, should `ul` and `li` be used to enforce the semantic?

src/lib/components/ListItemAction.svelte
```diff
@@ -0,0 +1,55 @@
...
+      display: flex;
+      align-items: center;
+      justify-content: flex-start;
+      gap: var(--padding-2x);
+
+      height: var(--item-height);
+    }
+
+    .icon-wrapper {
+      width: var(--item-height);
```
peterpeterparker, line 41: the width is the height?
or would `aspect-ratio` works well here?

src/lib/icons/IconLockOpen.svelte
```diff
@@ -1,9 +1,13 @@
 <!-- source: https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Alock_open%3A -->
+<script lang="ts">
+  export let size = `18px`;
```
peterpeterparker, line 3: the `DEFAULT_ICON_SIZE` is 20, it has to be 18?

src/routes/(split)/components/list-item-action/+page.md
```diff
@@ -0,0 +1,117 @@
...
+    <button slot="actions" class="secondary">One action</button>
+  </ListItemAction>
+</div>
+
+<style>
+  .wrapper {
+    background-color: var(--card-background);
+    margin: var(--padding-3x) 0;
+
+    & p {
```
peterpeterparker, line 93: out of curiosity, what's the purpose of `&` followed by a space?

src/routes/(split)/components/list-item-action/+page.md
```diff
@@ -0,0 +1,117 @@
...
+      <p>Unlocking</p>
+      <p class="description">No age bonus</p>
+    </div>
+    <button slot="actions" class="secondary">One action</button>
+  </ListItemAction>
+</div>
+
+<style>
+  .wrapper {
+    background-color: var(--card-background);
```
peterpeterparker, line 90: in NNS dapp, will it be use with a background too?

src/lib/components/ListItemAction.svelte
```diff
@@ -0,0 +1,55 @@
+<div class="container">
+  <div class="start-wrapper">
+    <div class="icon-wrapper">
+      <slot name="icon" />
```
peterpeterparker, line 4: the icon is mandatory or optional? if mandatory, should it be a property?

Just asking in case it would make the component usage easier?

peterpeterparker (commented):
Looks good, thanks. Few minor questions.

src/lib/components/ListItemAction.svelte
lmuntaner: Good question. I wasn't sure about it because I didn't enforce a parent that is `ul`.

How about adding the tagName as a prop?

src/lib/components/ListItemAction.svelte
lmuntaner: cool, yes! I didn't know about this rule

src/lib/icons/IconLockOpen.svelte
lmuntaner: I know. I didn't want to change the size in this PR. Just add the prop to configure it.

If we change this size, I guess we'll need to change others that go in the same place like the normal lock clock for example.

src/routes/(split)/components/list-item-action/+page.md
lmuntaner: No. This is just to highlight the element in the page.

src/lib/components/ListItemAction.svelte
peterpeterparker: I'm confuse now, does this component build the list and the items or only one item?

If only item, then I guess we can ignore my question but, maybe should we rename the component `ItemAction`?

src/lib/components/ListItemAction.svelte
peterpeterparker: Gotcha

src/lib/icons/IconLockOpen.svelte
peterpeterparker: Makes sense

src/routes/(split)/components/list-item-action/+page.md
peterpeterparker: I see, alright. I don't mind if you remove it because I was a bit unsure about the UI but, just thinking out at loud.

src/routes/(split)/components/list-item-action/+page.md
peterpeterparker: Ah, in that case, I think you can omit it.

This

```
.wrapper {
   & p {
      margin: 0;
    }
}
```

is the same as following I think:

```
.wrapper {
   p {
      margin: 0;
    }
}
```

P.S.: btw. just noticed that the `<style>` has no language `<style lang="scss">`, not sure if needed neither

peterpeterparker: But of course feel free to keep it if you rather like.

artkorotkikh-dfinity:
@lmuntaner from design standpoint this component works as expected on desktop and mobile :+1:

**Things to adjust:** 
Padding on mobile between content  and actions is too tight, san we use --padding-2x on mobile? 
image

**Questions:** 

1. Out of curiosity, when it comes to `` , how does the Unlocking icon look like in code? Is it a reusable icon from the lib (without a background) + css styling, or is it a separate svg that includes the background? 
2. Is it ok to assume that this component will support two buttons in one line?

src/lib/components/ListItemAction.svelte
lmuntaner: Just one item.

I see the confusion. I'll rename but I'll keep the tagName as prop. It can be useful because this component can be used inside lists.

lmuntaner:
@artkorotkikh-dfinity thanks for the review

I increased the gap between items in mobile:

Screenshot 2023-07-11 at 08 02 23

Is this ok?

Answers:

1. The Unlocking icon is only the icon in the middle of the background. So yes, it's reusable. We are currently using it in the NNS Dapp.
2. Correct, this component can support multiple buttons in the same line. Let me add it in the showcase.

src/routes/(split)/components/list-item-action/+page.md
lmuntaner: Ah, that's the markdown, I didn't know it also supported scss, great!

src/lib/components/ListItemAction.svelte
lmuntaner: done

lmuntaner:
@artkorotkikh-dfinity I also updated the description with a showcase with two buttons. Let me know what you think.

peterpeterparker (approved):
LGTM, thx

artkorotkikh:
@lmuntaner  Looks good to me too!

PR #235, opened 2023-07-10
build: bump dependencies

peterpeterparker:
# Motivation

Bump few dev dependencies and `dompurify`

# Changes

- bump dependencies
- run prettier after i18n

PR #238, opened 2023-07-10
build: downgrade Playwright

peterpeterparker:
# Motivation

Bumping Playwright in #235 has for effect to make the CI hangs on the E2E step.

Not the same version but, there seems to be an open issue in Playwright Repo: https://github.com/microsoft/playwright/issues/24071


lmuntaner (approved):
Thanks!

PR #240, opened 2023-07-11
Feat: Underline text and ghost button

lmuntaner:
# Motivation

Users have complained that the "ghost" and "text" buttons don't look like buttons at all.

Specially since the release of the "Select All" and "Clear" buttons for the filter modal.

# Changes

* Underline the text of the "ghost" and "text" buttons.

# Screenshots

Before:

Screenshot 2023-07-11 at 10 36 03

Now:

Screenshot 2023-07-11 at 10 36 52

### DIFF
--- a/src/lib/styles/global/button.scss
+++ b/src/lib/styles/global/button.scss
@@ -14,13 +14,13 @@ button {
     width: fit-content;
 
     @include fonts.standard;
+    text-decoration: underline;
 
     color: inherit;
 
     &:hover,
     &:active,
     &:focus {
-      text-decoration: underline;
       color: var(--value-color);
     }
 


lmuntaner:

peterpeterparker:
Makes sense but, we need an `hover` effect then.

lmuntaner:
> 

It changes the color. See the video below:

https://github.com/dfinity/gix-components/assets/4550653/95b0ccb3-e3df-4804-a72b-ced92a2c1dec

I also asked for Artem's review

peterpeterparker:
In light mode gonna be really subtle but, thanks for the feedback and video!

artkorotkikh-dfinity:
@peterpeterparker, @lmuntaner  I wouldn't make these buttons underlined, this will make them look like links (i.e. links usually open a new page). 

Can we just use a primary color for them? This will indicate that these elements are interactive. 

Not super relevant (sine we haven't implemented it) but example from Figma with "Max" button in the input:  
image

peterpeterparker:
Same style is applied for example in proposals (see screenshot), so personally, I would probably not change it even if some complained because it make sense to me the way they are 😅. Likewise I would personally maybe not make those `primary` because they aren't a primary call to action. But that's only my two cents, that's why I accepted the PR and fine with any other style, so don't mind me 😉.

Capture d’écran 2023-07-11 à 11 05 57

lmuntaner:
For example, let's remove the "Filter by ..." text in mobile:

Screenshot 2023-07-11 at 11 31 57

peterpeterparker:
> For example, let's remove the "Filter by ..." text in mobile:

Good idea, maybe even remove it everywhere.

And if you are chasing space and actoin, you could even merge "Select All" and "Clear". What I do in Juno, just one action for both. Not sure it's the best UX though.

demo

artkorotkikh-dfinity:
Ok, I'm not sure how to proceed with this. Do we just remove "Filter by ..." at this stage? 


I have a few suggestions for styles cleanup, but I don't think they relate to this specific issue. They can include:
- Updating secondary colors (to the more contrast ones) 
- We can leave Ghost button style as is, but I'd change at least a hover state to have secondary color without underline, something like this: 
image

lmuntaner:
> And if you are chasing space and actoin, you could even merge "Select All" and "Clear". What I do in Juno, just one action for both. Not sure it's the best UX though.

I always find it confusing to have the "Select All" or "Clear" as an option along the list. But maybe that's me. For example, if you select all, the unselect only one, should you also unselect the checkbox with "Select A"?

@artkorotkikh-dfinity the current hover state underlines the text:

https://github.com/dfinity/gix-components/assets/4550653/adb0790a-c4f6-4f50-a926-3e82d6df4c5d

We can change it to another color. Up to you. But I'll close this PR and we can take this in another time.

PR #242, opened 2023-07-11
build: vitest

peterpeterparker:
# Motivation

Replaces jest by vitest.

# PRs

- [x] #243
- [x] #244
- [x] #245

# Changes

- remove jest dependencies
- remove jest config
- rename `jest-setup.ts` to `vitest.setup.ts`
- add and integrate `vitest`
- configure `vitest.config.ts`
- update tests for vitest (remove comment about jsdom which is now default, replace `jest.` with `vi.` and few required tweaks)
- move flaky tests because of Svelte animation and transition to e2e tests

# Notes

I've try lots of things to fix the flaky tests for vitest but, really did not find any sustainable solutions. I understand the Svelte animation is now interpreted by vitest or jsdom but, what was really suprising is that I observed test taking more than 50 sec to even not comply. It looks like animation have side effects on rerender of components where store are updated in the test environment. Long story short: no better idea and using e2e, the tests are still covered as expected.


bitdivine (approved):
Looks cool to me.  You might want reviews from others as well though.

tsconfig.spec.json
```diff
@@ -1,4 +1,20 @@
...
+  },
+  "exclude": [],
+  "include": [
+    "ambient.d.ts",
+    "./types/**/$types.d.ts",
+    "./vite.config.ts",
+    "./src/**/*.js",
+    "./src/**/*.ts",
+    "./src/**/*.svelte",
+    "./tests/**/*.js",
```
mstrasinskis, line 14: Does `.(js|ts|svelte)` syntax works here?

mstrasinskis (approved):
Great work, thanks!

PR #245, opened 2023-07-12
feat: busy store type and reset

peterpeterparker:
# Motivation

Add types to busy store and a reset function for test purpose.


src/lib/stores/busy.store.ts
```diff
@@ -29,6 +40,10 @@ const initBusyStore = () => {
         state.filter(({ initiator }) => initiator !== initiatorToRemove)
       );
     },
+
+    reset() {
```
mstrasinskis, line 44: In `nns-dapp` some stores have `resetForTesting()`. What do you think about having the same name, to make its role more obvious?

mstrasinskis (approved):
Thank you!

src/lib/stores/busy.store.ts
peterpeterparker: Makes sense, thanks! Done.

PR #246, opened 2023-07-12
feat: set transitions as local

peterpeterparker:
# Motivation

Svelte transitions are executed by default in components regardless if their parent components destroy those or not. This behavior is known as the `|global` transitions setting and is the default in Svelte v3.

This leads to issue with the SvelteKit navigation because DOM elements might not be detached and related context might remain active.

We generally notice the issue in NNS dapp after navigation when the DOM wrongly contains two `split-pane` elements.

In Svelte v4 the default behavior was changed to `|local`. Setting that can already be used in v3 when explicitely set. This setting has for effect to not execute the transition and destroy the components if parents destroy those. i.e. to populate the destroy.

# References

- documentation: https://svelte.dev/docs/v4-migration-guide#transitions-are-local-by-default
- svelte issue: https://github.com/sveltejs/svelte/issues/6686

# Changes

- scope all transitions of Svelte to `|local`

### DIFF
--- a/src/lib/components/Backdrop.svelte
+++ b/src/lib/components/Backdrop.svelte
@@ -17,8 +17,8 @@
   role="button"
   tabindex="-1"
   aria-label={$i18n.core.close}
-  in:fade={{ duration: FADE_IN_DURATION }}
-  out:fade={{ duration: FADE_OUT_DURATION }}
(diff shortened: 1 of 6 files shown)


mstrasinskis (approved):
Let's try! Thanks!

PR #258, opened 2023-07-26
Feat: ItemAction - do not stretch actions

lmuntaner:
# Motivation

The ItemAction should not stretch the actions slot.

# Changes

* Remove `stretch` style in the actions slot.

# Screenshots

No UI changes in desktop.

New UI in mobile:
Screenshot 2023-07-26 at 09 19 43

### DIFF
--- a/src/lib/components/ItemAction.svelte
+++ b/src/lib/components/ItemAction.svelte
@@ -10,7 +10,7 @@
     </div>
     <slot />
   </div>
-  <div class="actions">
+  <div>
     <slot name="actions" />
   </div>
 </svelte:element>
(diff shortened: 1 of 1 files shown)


lmuntaner:

src/lib/components/ItemAction.svelte
```diff
@@ -10,7 +10,7 @@
     </div>
     <slot />
   </div>
-  <div class="actions">
+  <div>
```
peterpeterparker, line 13: you still need the `div` then?

lmuntaner: Yes. I want the spacing to be set by the container here. If I add a slot and the slot is two elements, the gap here will apply also between those two elements. Which is not what we want.

peterpeterparker (approved):
LGTM, thanks

PR #262, opened 2023-07-31
feat: true card links

peterpeterparker:
# Motivation

Currently setting the `role=link` to the card only set the role but, does not make the card behaves like a true link. To improve the usability and navigation in NNS dapp, it is interesting to transform this behavior and render real links.

# Changes

- deprecate role `link` (replaced) and `radio` (not use anymore)
- introduces property `href`
- render a link if `href` else render an `article`
- block click propagation if `href` is set
- update documentation to use links


src/lib/components/Card.svelte
```diff
@@ -42,12 +43,24 @@
...
   })();
+
+  const dispatch = createEventDispatcher();
+
+  const onClick = ($event: MouseEvent | KeyboardEvent) => {
+    if (clickable && nonNullish(href)) {
+      return;
+    }
+
+    dispatch("click", $event?.detail);
```
mstrasinskis, line 54: Nice.

mstrasinskis (approved):
Thanks!

PR #261, opened 2023-07-31
feat: maintain xlarge layout on large screen

peterpeterparker:
# Motivation

Artem suggested to keep the menu open and display the content on large screen as if it was still extra large screen.

# Changes

- transform xlarge media breakpoints to large
- remove card icon (the arrow on the proposal page lead to issue, discussed with Artem)
- update `<Island />` to fit the appropriate width
- adapt grid to better fit the space

# Screenshots

Mainnet:

a1
a2
a3
a4
a5

PR:

b1
b2
b3
b4
b5

PR #263, opened 2023-08-01
feat: revert two icons in Card

peterpeterparker:
# Motivation

Turns out we still need two icons in the `Card` component because those are used in the universe selector of NNS dapp.

# PRs

- [x] Revert partially https://github.com/dfinity/gix-components/pull/261


mstrasinskis (approved):
LGTM, thanks!

PR #264, opened 2023-08-02
fix: showcase style workaround

peterpeterparker:
# Motivation

For simplicity reason we do not have yet implemented the components picker in the showcase but just a style workaround. Because the card has become links, a selector should be updated.

# Screenshots

Capture d’écran 2023-08-02 à 09 56 03


mstrasinskis (approved):
LGTM, thanks!

PR #265, opened 2023-08-02
fix: card link spacing with inherited margin

peterpeterparker:
# Motivation

I discovered an issue with the spacing due to the new card links in NNS dapp that can only be reproduced by using the packaged gix-cmp. If the repo is installed as reference, the issue does not surfaces.

# Changes

- set margin size for card displayed as links

# Screenshots

Capture d’écran 2023-08-02 à 10 24 46


mstrasinskis (approved):
Thank you!

PR #270, opened 2023-08-07
style: card link hover color

peterpeterparker:
# Motivation

Not really an issue when the all card is a text but, when only parts of it has no specific color, hover a card link has for effect to change the color which does not look that good on the NNS dapp launchpad.

# Screenshots

hover

### DIFF
--- a/src/lib/components/Card.svelte
+++ b/src/lib/components/Card.svelte
@@ -86,7 +86,8 @@
   @use "../styles/mixins/display";
   @use "../styles/mixins/card";
 
-  article, a {
+  article,
+  a {
     display: flex;
     flex-direction: column;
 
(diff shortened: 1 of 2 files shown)


src/lib/components/Card.svelte
```diff
@@ -166,6 +166,10 @@
 
   a.card {
     margin: 0;
+
+    &:hover {
```
mstrasinskis, line 170: Will it override disabled state?

mstrasinskis (approved):
LGTM, thanks!

src/lib/components/Card.svelte
peterpeterparker: Oh excellent point and well spotted!!!

Modified the style to:

```
&:not(.disabled):hover {
      color: inherit;
    }
```

PR #271, opened 2023-08-07
fix: prevent default behavior of copy button

peterpeterparker:
# Motivation

When the `Copy` component is used within an hyperlinks, as e.g. a card, it triggers both its copy action but also propagate the default behavior which has for effect to trigger the link.


lmuntaner (approved):
Thanks!

PR #272, opened 2023-08-09
Chore: Improve Section when no description

lmuntaner:
# Motivation

There is too much spacing when there is no description in the Section.

# Changes

* New class to change margins when description slot is not present.

# Screenshots

Before:
Screenshot 2023-08-09 at 07 31 40

Now:
Screenshot 2023-08-09 at 07 31 46

Now with and without:
Screenshot 2023-08-09 at 07 29 35

### DIFF
--- a/src/lib/components/Section.svelte
+++ b/src/lib/components/Section.svelte
@@ -1,14 +1,17 @@
 <script lang="ts">
   export let testId: string | undefined = undefined;
+
+  let noDescription = false;
+  $: noDescription = $$slots.description === undefined;
 </script>
 
 <div class="container" data-tid={testId}>
-  <div class="section-title">
+  <div class="section-title" class:noDescription>
     <slot name="title" />
     <slot name="end" />
   </div>
   <slot name="description" />
-  <div class="content-wrapper"><slot /></div>
+  <div class="content-wrapper" class:noDescription><slot /></div>
 </div>
 
 <style lang="scss">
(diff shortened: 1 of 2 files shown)


lmuntaner:

src/lib/components/Section.svelte
```diff
@@ -26,9 +29,17 @@
     justify-content: space-between;
 
     margin-bottom: var(--padding);
+
+    &.noDescription {
```
peterpeterparker, line 33: You can spare one assignment here I think

```
&:not(.noDescription) {
  margin-bottom: var(--padding);
}
```

peterpeterparker (commented):
generally speaking, have you try using `row-gap` on the `.container` instead of defining margin for each element and adding an additional class `.noDescription`?

Wouldn't following do the trick?

```
.container {
    width: 100%;

    display: flex;
    flex-direction: column;
    align-items: stretch;
    row-gap: var(--padding-2x);
  }
```

peterpeterparker:
Ok, as you wish.

Regarding the other comment https://github.com/dfinity/gix-components/pull/272#discussion_r1287993061 ?

lmuntaner:
> Ok, as you wish.
> 
> Regarding the other comment [#272 (comment)](https://github.com/dfinity/gix-components/pull/272#discussion_r1287993061) ?

It's not only as I wish, right? I can't use the gap if they are different.

I'll add the change in the comment, I forgot to reply, thanks!

src/lib/components/Section.svelte
lmuntaner: thanks!

peterpeterparker (approved):
LGTM, thx

PR #277, opened 2023-08-20
feat: menu collapse

peterpeterparker:
# Motivation

Introduce an option to collapse or extend the menu on large+ screen.

# Notes

The solution proposed is implemented with a global CSS first approach - identitical to what we do for the theme - to avoid loading glitch when we will deploy it on the IC. That way the style does not need to wait for the JS to be loaded but can be presented in the correct state as soon as the CSS is loaded.

# Changes

- add "Collapse / expand" button in menu
- globally style the button - `menu.scss` - including hiding or displayed the button according state
- new store to hold the "collapsed" menu state `menu.store.ts`
- as for the theme, the state is reflected to the local storage
- snippet to load the menu state in `index.html`
- add option `title` to `MenuItem`

# Screenshots

See PR tests results or https://github.com/dfinity/nns-dapp/pull/3153


src/app.html
```diff
@@ -125,6 +125,25 @@
       }
     </script>
 
+    <!-- Init menu as fast as possible as well -->
+    <script>
+      try {
+        const currentMenu =
+          localStorage.nnsMenu === null ||
```
mstrasinskis, line 132: nitpick: Just because it's a pure js, how about `!localStorage.nnsMenu ? undefined ...`? Just to preserve some bits.

src/lib/styles/global/menu.scss
```diff
@@ -0,0 +1,105 @@
...
+
+          width: var(--padding-4x);
+          height: var(--padding-4x);
+          min-height: var(--padding-4x);
+
+          transform: rotate(0deg);
+
+          display: none;
+
+          @media (min-height: 654px) {
```
mstrasinskis, line 31: Is this some empirical value?

mstrasinskis (approved):
LGTM, just a couple of questions, thanks!

src/app.html
peterpeterparker: Good idea! I copied the code we have for the `localstorage.nnsTheme` but, that's a good idea, let's do it.

peterpeterparker: Done.

src/lib/styles/global/menu.scss
mstrasinskis: Asking because I saw this value in nns-dapp too.

peterpeterparker: Yes. I added comments now.

mstrasinskis: And if it can be a css variable that could be nice. If not/not-elegant, then that's not a problem either.

PR #280, opened 2023-08-24
Fix: Show svg icon in Safari

lmuntaner:
# Motivation

The svg IconXxxPage was not appearing in Safari.

The problem is that the width and height were not defined in the svg and Safari needs it to be defined.

# Changes

* Set a default size of 100% in the svgs so that the parent controls the size.

# Screenshots

Both screenshots of Safari.

Screenshot 2023-08-24 at 08 08 37
Screenshot 2023-08-24 at 08 09 42


peterpeterparker (approved):
LGTM, thx

PR #282, opened 2023-08-24
Fix: PageBanner for Safari

lmuntaner:
# Motivation

The PageBanner as is couldn't go directly inside a flex container because of a height overflow in Safari.

More on the issue: https://stackoverflow.com/questions/62075401/safari-grid-in-flexbox-produces-height-overflow

# Changes

The fix is to add a wrapper around the container of PageBanner.

# Screenshots

No changes.

### DIFF
--- a/src/lib/components/PageBanner.svelte
+++ b/src/lib/components/PageBanner.svelte
@@ -2,17 +2,22 @@
   export let testId: string | undefined = undefined;
 </script>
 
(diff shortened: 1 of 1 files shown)


lmuntaner:

I tested it in NNS Dapp and the wrapper is indeed not needed anymore.

peterpeterparker (approved):
Thx

PR #283, opened 2023-08-24
GIX-1836: Add new primitive colors

lmuntaner:
# Motivation

We want to synchronize the colors in Figma and NNS Dapp.

In this PR, we introduce the color primitive used in Figma, but we don't use them yet.

Source of colors: https://www.figma.com/file/P4JbeFs6oGt2WmKeB4PHFc/Updated-NNS-Styles-%26-Components?type=design&node-id=514-220550&mode=design&t=4slnxbCcIfhrcgu6-0

Under Local Variables > Primitives

# Changes

* New styles global file color.scss
* Add new file to global imports.

# Screenshots

Nothing changed yet.


lmuntaner:

src/lib/styles/global/color-primitives.scss
```diff
@@ -0,0 +1,50 @@
...
+  --purple-dark-400: #483666;
+  --purple-dark-450: #3d2a5c;
+  --purple-dark-500: #372652;
+  --purple-dark-550: #31204c;
+  --purple-dark-600: #2d1a4a;
+  --purple-dark-650: #291942;
+  --purple-dark-900: #24143d;
+
+  // Accessory
+  --fade: rgba(36, 20, 61, 0.6); // #24143d 60%
```
mstrasinskis, line 32: What is this colour for? I didn't find it in Figma.

mstrasinskis (approved):
LGTM, thanks!

src/lib/styles/global/color-primitives.scss
mstrasinskis: If there will be some derived colours, maybe `--fade` should live there?

mstrasinskis: `rgba(var(--purple-dark-900), 0.6)`?

peterpeterparker (approved):
LGTM, thanks.

A suggestion if you would like it: `color-primitives.scss` can also be shortened to `colors.scss`

lmuntaner:
> LGTM, thanks.
> 
> A suggestion if you would like it: `color-primitives.scss` can also be shortened to `colors.scss`

I wasn't sure with the name, but I also prefer it shorter.

src/lib/styles/global/color-primitives.scss
lmuntaner: I don't know, but it's a good question. I'll ask it to Artem and then change based on the feedback.

The derived colors will be in each theme. I will ask him where we should put those. I agree that maybe this one belongs there if it's a derived color.

PR #284, opened 2023-08-25
GIX-1789: New colors

lmuntaner:
# Motivation

We want to align the colors in Figma and the NNS Dapp.

In this PR, we use the colors used in Figma for the theme variables.

More work is needed to align, clean and improve the variable colors.

# Changes

* Secondary button doesn't need the trick to have a border with gradient. Therefore, we remove the extra complexity because we can use a transparent background.
* Add three new primitive colors.
* Change all primitive color values in the theme files light.scss and dark.scss to use the new colors.

# Screenshots

NNS Dapp deployed here https://kbruq-4yaaa-aaaaa-aacta-cai.nnsdapp.dfinity.network/ for testing.


src/lib/components/Toast.svelte
```diff
@@ -27,7 +27,7 @@
       ["error"]: IconError,
       ["info"]: IconInfo,
       ["custom"]: undefined,
-    }[level]);
+    })[level];
```
peterpeterparker, line 30: is this correct? not sure how it relates to the PR

src/lib/components/Toast.svelte
```diff
@@ -38,6 +38,7 @@
   let overflow: "scroll" | "truncate" | "clamp" | undefined;
   let position: ToastPosition | undefined;
   let icon: typeof SvelteComponent | undefined;
+  // TODO-colors: Can we remove this?
```
peterpeterparker, line 41: probably no because of the toast top right that sometimes has to be forced in a particular theme

peterpeterparker (approved):
LGTM, thanks

Added two minor comments if you want to have a look before merge.

src/lib/components/Toast.svelte
lmuntaner: thanks for the tip!

src/lib/components/Toast.svelte
lmuntaner: It was my lint that changed it.

PR #289, opened 2023-09-04
build: copy workers after build instead of postinstall

peterpeterparker:
# Motivation

When we locally install gix-cmp in NNS dapp

```
$ npm install /Users/dskloet/dev/gix-components
```

the installation runs also the pre/post scripts of gix-cmp. As I added a post scripts to copy the web workers for the analyytics of the showcase, this script gets executed which is not necessary.

So to avoid this issue and because we only needs the web workers for the showcase, we can move the script from `postinstall` to a manually call after build.

# Changes

- move script from `postinstall` to `build`
- do not print error in script if no error

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "license": "Apache-2.0",
   "scripts": {
     "dev": "npm run i18n && vite dev",
-    "build": "npm run i18n && vite build",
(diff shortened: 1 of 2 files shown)


dskloetd (approved):
Thanks!

PR #291, opened 2023-09-05
build: fix copy workers

peterpeterparker:
# Motivation

#289 moved the copy from the workers from `postinstall` to a post build script but, since `static` files are moved to `build` output folders, the post script that copy to `static` has no effect. That's why the script should directly copy the files to the output.


mstrasinskis (approved):
Rubber stamp.

PR #292, opened 2023-09-05
Stop QRCodeReader if component destroyed before started

dskloetd:
# Motivation

Currently if the QRCodeReader component is destroyed before the scanner is started, it fails to stop the scanner and then the camera is turned on and stays on until the page is reloaded.

# Changes

1. Don't try to stop the scanner if it's not started.
2. Stop the scanner after starting if the component is destroyed.

# Screenshots

There is still an error if they component is destroyed because the reader is started but at least the camera won't remain on.

```
html5-qrcode.js?v=a11cfc36:27301 Uncaught (in promise) TypeError: Failed to execute 'removeChild' on 'Node': parameter 1 is not of type 'Node'.
    at html5-qrcode.js?v=a11cfc36:27301:23
    at async QRCodeReader.svelte?t=1693908189579:105:4
qr-code-reader:1 Uncaught (in promise) DOMException: The play() request was interrupted because the media was removed from the document. https://goo.gl/LdLk22
```

### DIFF
--- a/src/lib/components/QRCodeReader.svelte
+++ b/src/lib/components/QRCodeReader.svelte
@@ -10,6 +10,9 @@
 
   let html5QrCode: Html5Qrcode | undefined;
 
+  let ScannerState: Record<string, number>;
+  let isDestroyed = false;
+
   onMount(async () => {
     const qrCodeSuccessCallback = (decodedText: string) =>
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
LGTM, thanks

PR #293, opened 2023-09-11
GIX-1864: Add property noPadding on Card

lmuntaner:
# Motivation

Fix the poor contrast on the token selector for ICP logo described here: https://dfinity.atlassian.net/browse/GIX-1864

Solution: add a stronger outline to the logo when selected and on hover.

Yet, to be able to know when the card is hovered, we can't rely on the hover of the content because the Card has some padding. Therefore, I'd like to add a `noPadding` prop to the Card component. This way, the child will manage the padding and can also manage the onhover of the whole Card.

# Changes

* Add prop `noPadding` in `Card` component.

# Screenshots

Screenshot 2023-09-12 at 09 30 51

### DIFF
--- a/src/lib/components/Card.svelte
+++ b/src/lib/components/Card.svelte
@@ -57,6 +57,8 @@
   {role}
   data-tid={testId}
   on:click={onClick}
+  on:mouseenter
+  on:mouseleave
   class={`card ${theme ?? ""}`}
   class:clickable
   class:icon={nonNullish(icon)}
--- a/src/lib/components/Card.svelte
+++ b/src/lib/components/Card.svelte
@@ -14,6 +14,7 @@
   export let theme: "transparent" | "framed" | "highlighted" | undefined =
     undefined;
   export let href: string | undefined = undefined;
+  export let noPadding = false;
 
   let container: "article" | "a" = "article";
   $: container = nonNullish(href) ? "a" : "article";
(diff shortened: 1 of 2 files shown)


lmuntaner:

peterpeterparker:
> add a stronger outline to the logo when selected and on hover.

Why not solving this with CSS? `:hover` and `:active`?

lmuntaner:
> Why not solving this with CSS? :hover and :active?

Where do you propose to add them? I tried in the content of the Card, but if the user goes over the card, but not over the content, it doesn't detect the `:hover`.

Screenshot 2023-09-11 at 11 08 16

peterpeterparker:
With a `:global` maybe?

peterpeterparker:
Yeah but using JavaScript code to hack around what CSS does natively isn't optimal neither and less performant. That's why I emitted the idea.

Don't have a better idea though.

Does it really need a particular color on hover? In previous version, before reviewing the colors, we managed to have one border or shadow that worked for both use case.

lmuntaner:
> 
> Does it really need a particular color on hover? In previous version, before reviewing the colors, we managed to have one border or shadow that worked for both use case.

Seeing how ugly it is to change the border, I agree we should challenge the design first. I'll talk with Artem.

lmuntaner:
@peterpeterparker ready for a review

peterpeterparker (approved):
LGTM, thanks.

Just in case you would like the idea, we can also use a CSS variable. If used with Svelte, then svelte adds a div automatically in the dom with display no contents.

```
// Card component

<style>
   article
       --padding: var(--card-padding, var(--padding-the-default-value);
       
// NNS dapp
<Card --padding="0">
```

(pseudo code)

PR #294, opened 2023-09-12
Chore: Remove unused text2 color

lmuntaner:
# Motivation

Color `--text2-color` was only set in dark mode and that caused an issue that was solved https://github.com/dfinity/nns-dapp/pull/3306

In this PR, this color is removed and the two references are substituted with the primitive color.

This has been agreed with Artem.

# Changes

* Remove `--text2-color` from dark colors and change references to primitive color.

# Screenshots

No changes. It wasn't used in any component.


peterpeterparker (approved):
LGTM, thx

PR #295, opened 2023-09-14
build: bump analytics

peterpeterparker:
# Motivation

There were major improvements in the Juno's analytics canister to improve scalability of the data. Those changes lead to few structure changes in the keys and entities which are reflected in the last version of the library.

So we need to bump the library to reactive the analytics for gix-cmp.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
     "LICENSE"
   ],
   "devDependencies": {
-    "@junobuild/analytics": "^0.0.5",
     "@playwright/test": "^1.32.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -65,6 +64,7 @@
     "vitest": "^0.33.0"
   },
   "dependencies": {
+    "@junobuild/analytics": "^0.0.9",
```
dskloetd, line 67: Why is it no longer a `devDependencies`?

peterpeterparker: No particular reason. Redo.

dskloetd (approved):

PR #296, opened 2023-09-21
GIX-1918: Add segments to ProgressBar

lmuntaner:
# Motivation

ProgressBar handles different progress segments to show in different colors.

Keep previous props `value` and `color` for backwards compatibility.

We don't want to remove them because they offer a simpler and more readable interface for the common use case.

# Changes

* Add new prop `segments` from type `Segment` to ProgressBar.

# Screenshots

progress-bar-multiple-colors


lmuntaner:

src/lib/components/ProgressBar.svelte
```diff
@@ -4,16 +4,54 @@
   export let max = 1;
   export let value: number;
   export let color: "warning" | "primary" = "primary";
+  type Segment = { value: number; color: string };
```
peterpeterparker, line 7: shouldn't we expose this type?

peterpeterparker (approved):
LGTM, thanks

Note: we assume the consumer pass a total of segments' value that is correct, equals 100%, I guess we do not need to overengineer here

src/lib/components/ProgressBar.svelte
lmuntaner: Yes, makes sense. Let me move it.

PR #297, opened 2023-09-21
Fix: Rename Segment type due to conflict

lmuntaner:
# Motivation

"Segment" is already a component. Therefore, I renamed the type to ProgressBarSegment.

# Changes

* Rename type Segment to ProgressBarSegment.

# Screenshots

No changes.


peterpeterparker (approved):
LGTM

PR #298, opened 2023-09-22
Fix: `value` prop optional in ProgressBar

lmuntaner:
# Motivation

The `value` prop in the ProgressBar component is not used if there are any `segments`. Therefore, it shouldn't be mandatory.

# Changes

* Add a default value of `0` to the `value` prop in ProgressBar.
* Update documentation.
* Not related, but added a `testId` prop to ProgressBar as well.

# Screenshots

No changes in the component UI. Just the interface and the docs.


lmuntaner:
@dskloetd please review

src/lib/components/ProgressBar.svelte
```diff
@@ -4,7 +4,9 @@
   // Html default is 1 anyway
   // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress?retiredLocale=ca#attr-max
   export let max = 1;
-  export let value: number;
+  // `value` becomes now optional
```
dskloetd, line 7: This comment was useful for the review but not worth keeping, right?
Otherwise you might as well write on each of the props that they are optional. Or am I missing something?

lmuntaner: I initially thought of leaving it. But yes, it adds no value.

lmuntaner:

dskloetd (approved):

PR #300, opened 2023-10-03
Chore: Fix vulnerability issues

lmuntaner:
# Motivation

Fix security issues.

# Changes

Only automatic changes after running `npm audit fix`

# Screenshots

No changes.

PR #301, opened 2023-10-03
build: bump vitest

peterpeterparker:
# Motivation

Bump vitest and vite-plugin-svelte (because of issue https://github.com/dfinity/gix-components/pull/300#issuecomment-1744910316 which was solved in https://github.com/sveltejs/vite-plugin-svelte/issues/710#issuecomment-1672661958).


lmuntaner (approved):
Thanks!

PR #302, opened 2023-10-04
build: fix running e2e locally with staging build env

peterpeterparker:
# Motivation

Fix e2e error `[WebServer] Error: Not found: /components/workers/analytics.worker.js` by introducing a staging mode that can be use to build and run the test in order to avoid loading the analytics.

# Changes

- add mode `stagin`
- build test with staging instead of build
- don't load anyltics on staging


mstrasinskis (approved):
LGTM, thank you!

PR #306, opened 2023-10-06
Add `content-cell-disabled-island` style

mstrasinskis:
# Motivation

Add a utility class for sub/alternative islands.

# Changes

- add `content-cell-disabled-island` style

# Screenshots

image

### DIFF
--- a/src/lib/styles/global/utility-classes.scss
+++ b/src/lib/styles/global/utility-classes.scss
@@ -46,6 +46,13 @@
   border-radius: var(--border-radius);
 }
 
+.content-cell-disabled-island {
+  @include content.disabled-island;
+
+  padding: var(--padding-2x);
+  border-radius: var(--border-radius);
+}
+
 // Spacing
 
 .no-margin {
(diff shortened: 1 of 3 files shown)


src/lib/styles/global/utility-classes.scss
```diff
@@ -46,6 +46,13 @@
   border-radius: var(--border-radius);
 }
 
+.content-cell-disabled-island {
```
lmuntaner, line 49: Why "disabled"? It looks more highlighted to me.

Do we really need to do all this with the "content-cell..." utility classes?

Are we using them only for proposals?

mstrasinskis: Yes, currently on for proposal summary. Why not, let me do it directly in nns-dapp.

lmuntaner: I'm ok to move it. Maybe we just need components shared between NNS and SNS proposals with the CSS inside them.

mstrasinskis: Currently, the ProposalSummary component is used for both. I suggest thinking about it when we need to reuse these styles.

mstrasinskis:
Not necessary anymore

PR #307, opened 2023-10-06
GIX-1938: Change Tag style

lmuntaner:
# Motivation

Adapt Tag component to new UI and design requirements.

# Changes

* Change Tag UI.
* Add new props "intent", "size", "testId" to Tag.
* Remove the CSS variable `--tag-color` for Tag. Instead it uses the "intent" prop.
* Change Tag page to explain new props and add examples.
* Four new CSS colors. These already follow the way Figma uses CSS variables for colors.

# Screenshots

Screenshot 2023-10-06 at 14 23 05
Screenshot 2023-10-06 at 14 26 54


src/lib/components/Tag.svelte
```diff
@@ -1,21 +1,56 @@
 <script lang="ts">
-  export let tagName: "span" | "h3" | "li" = "span";
+  export let tagName: "span" | "li" = "span";
+  export let intent: "warning" | "success" | "error" | "info" = "info";
```
mstrasinskis, line 3: nitpick: not a strong opinion but how about not having a default value for the intent? Otherwise developer should be aware of it.

mstrasinskis (approved):
Thanks, just a couple of minor comments.

peterpeterparker:
It's "Big Success" but "Large with icon". Should the text capitalization be defined with CSS?

src/lib/components/Tag.svelte
```diff
@@ -1,21 +1,56 @@
...
+    
+    // Force fit-content for `li` tags.
+    width: fit-content;
+    padding: var(--padding-0_5x) var(--padding);
 
-    color: var(--tag-color, var(--tag-default-color));
-    border: 1px solid var(--tag-color, var(--tag-default-color));
-    border-radius: var(--border-radius-5x);
+    &.large {
+      font-size: var(--font-size-standard);
```
peterpeterparker, line 25: There isn't a mixin to use the font standard?

src/lib/components/Tag.svelte
```diff
@@ -1,21 +1,56 @@
...
+    }
 
-    height: fit-content;
+    &.medium {
+      font-size: var(--font-size-small);
```
peterpeterparker, line 29: ditto

src/lib/components/Tag.svelte
```diff
@@ -1,21 +1,56 @@
...
+    }
+
+    &.info {
```
peterpeterparker, line 32: shouldn't be this de default?

src/lib/styles/themes/dark.scss
```diff
@@ -1,6 +1,14 @@
 /** Default theme: dark */
 
 :root {
+  // -------- NEW CSS VARIABLES ----------------
+  // We want to migrate to a new set of CSS variables for colors more adapted to our Figma.
+  --elements-divider: var(--purple-dark-400);
+  --text-description: var(--purple-dark-200);
+  --text-light: var(--purple-50);
+  --text-text: var(--purple-900);
```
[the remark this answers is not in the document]
lmuntaner: Yes, they have been aligned with Artem

src/lib/components/Tag.svelte
lmuntaner: Good point, I wanted to keep backwards compatibility though. So that we don't have to change the code.

lmuntaner:
> It's "Big Success" but "Large with icon". Should the text capitalization be defined with CSS?

No, I don't think so.

src/lib/components/Tag.svelte
lmuntaner: I wanted to be explicit.

src/lib/components/Tag.svelte
lmuntaner: I'm not sure I want to change the line height. Let me ask Artem.

src/lib/components/Tag.svelte
peterpeterparker: Yeah but then if comes with a tradeoff if anyone has ever the idea to make the property optional - it might loose it style.

src/lib/components/Tag.svelte
lmuntaner: I think it makes sense to change the line-height. I'll use the mixin.

src/lib/components/Tag.svelte
lmuntaner: done

src/lib/components/Tag.svelte
lmuntaner: done

lmuntaner:
@peterpeterparker ready for another review

peterpeterparker (approved):
LGTM, thx

PR #309, opened 2023-10-12
Key value info align icon right

mstrasinskis:
# Motivation

Add property `alignIconRight` to align info icons to the right

# Changes

- alignIconRight
- align value text right to improve long text alignment (mobile screenshot)

# Screenshots

## Mobile
image

## Desktop
image

### DIFF
--- a/src/lib/components/KeyValuePair.svelte
+++ b/src/lib/components/KeyValuePair.svelte
@@ -35,5 +35,6 @@
     display: flex;
     align-items: center;
     margin-inline-start: 0;
+    text-align: right;
   }
 </style>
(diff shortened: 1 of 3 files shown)


peterpeterparker:
What happens if the value is for example a copy button? Is that expected?

Text | Value button button

lmuntaner (approved):
LGTM!

If the user of the component puts an icon there, it's up to them to find a better solution, no?

The library can't fix all the misuses of the components.

peterpeterparker:
Well this issue cannot happen currently with this or any other component of the library no?

peterpeterparker:
Discussed offline with @mstrasinskis. Don't mind me.

PR #320, opened 2023-10-23
fix: layout-bottom-offset wrong value

mstrasinskis:
# Motivation

Remove unexpected border on proposal detail page by fixing the bottom-offset calculation.

# Changes

- new global variable `--content-margin` 
- update `--layout-bottom-offset` calculation in `Content.svelte`
Note: this variable is also used by `SplitContent` and `Toast` but it's always `0`.

# Screenshots

## Before
image

## After
before

after

### DIFF
--- a/src/lib/components/Content.svelte
+++ b/src/lib/components/Content.svelte
@@ -16,7 +16,7 @@
 
 <div
   class="content"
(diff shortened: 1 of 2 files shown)


peterpeterparker (approved):
Nice. Thanks!

PR #321, opened 2023-10-23
add a CSS reset

meodai:
# Motivation

To make sure all browser behave the same and no default element spaces need to be reseted all the time.

# Changes

Add a css reset.

Choose to track the reset in our own codebase, since we don't use `box-sizing: border-box;`


AntonioVentilii:

Thanks for the contribution!

PR #327, opened 2023-11-07
Add collapse all button

mstrasinskis:
# Motivation

Add icon for collapse all action.

# Changes

- `IconCollapseAll` component

# Screenshots

image


mstrasinskis:
> LGTM, thanks
> 
> Side note: it would be cooler to have a single icon that has a visual transition / animation between the two states collapse and not collapsed. Generally speaking I think it would be neat to add more visual small animations to the dapp


Thanks for your feedback. Let's start with the in:fade first :)

PR #329, opened 2023-11-08
Chore: Export breakpoint constants

lmuntaner:
# Motivation

Export the rest of the breakpoints as JS constants.

# Changes

* Add `BREAKPOINT_SMALL` and `BREAKPOINT_MEDIUM` to `constants.ts`.

# Screenshots

No changes.

### DIFF
--- a/src/lib/constants/constants.ts
+++ b/src/lib/constants/constants.ts
@@ -1,5 +1,9 @@
 export const DEFAULT_ICON_SIZE = 20;
 
+// 576 is the CSS $breakpoint-small size
+export const BREAKPOINT_SMALL = 576;
+// 768 is the CSS $breakpoint-medium size
+export const BREAKPOINT_MEDIUM = 768;
 // 1024 is the CSS $breakpoint-large size
 export const BREAKPOINT_LARGE = 1024;
 // 1300 is the CSS $breakpoint-extra-large size


src/lib/constants/constants.ts
```diff
@@ -1,5 +1,9 @@
 export const DEFAULT_ICON_SIZE = 20;
 
+// 576 is the CSS $breakpoint-small size
+export const BREAKPOINT_SMALL = 576;
```
peterpeterparker, line 4: Why do you need those?

`BREAKPOINT_LARGE` is unfortunately exposed because we had no other choice. We should retrain to use JS for styling purpose therefore asking.

lmuntaner: In the Tokens page, the UI is completely different for mobile and desktop. In one, I'll use a `table`, but in the other one I can't use a `table`.

See Figma: https://www.figma.com/file/x8dk5axQUf8Bi43uUsqCHc/NNS-Production-%26-Hand-over?type=design&node-id=408-2786&mode=design&t=xlsWY5h6H3iG9UPF-0

Therefore, I want the page to check the innerWidth of the window (like we do in `LayoutList`) and render one component or another depending on the window's width.

I don't need the small, only the medium breakpoint if you prefer that I remove one.

peterpeterparker: Wouldn't it work if you use a `div` with `role=table` and style according media breakpoint?

peterpeterparker: Which seems also more maintable because if we use JavaScript to render this or that then it means that the connection between the UI and the source of data will be duplicated.

lmuntaner: But I'm afraid that if we don't use a `<table>` then we won't be able to align columns. This will most probably come later that a third column will be added and will need to be aligned with the others.

Doing that without a `<table>` is too complex and that's why I prefer managing two components.

The logic won't be in this components, they will trigger events handled by the parents. So no logic will be duplicated.

I'm afraid that I use now a `<li>` for them, change styles based on breakpoint, and then when we need to aling vertically rows, we can't and we need to build a new component.

I see two as more maintainable than one with breakpoints and different CSS.

peterpeterparker: I don't think that's a valid concern. Semantically and styling wise a table layout can be achieved.

peterpeterparker: I am agree on the concern about the fact that it might become tricky at some point which can be validated with a POC. Did you try to follow this path before deciding implementing a table and using javascript to toggle between components?

lmuntaner: Not in Tokens in Figma yet, but you can see a similar table being used for proposals and neurons.

https://www.figma.com/file/29OdZHbNaOk9RkrfESucjb/NNS-1.03-WIP?type=design&node-id=3034-134590&mode=design&t=p9wsAM2MwEs7dvq3-0

I didn't want to create a reusable table yet, but that's the idea.

peterpeterparker: I still don't think it's a valid reason. Duplicating components as signicative cost both in term of performance and maintanance so unless it's really mandatory, I personally always try to avoid it. So far, not concinved it's mandatory.

peterpeterparker: Of course your and team call, just my two cents.

dskloetd: If it can be done in pure CSS, I think we should prefer that.
If Llorenc can't get it to work, maybe you can help in figuring out how to do it, David?

peterpeterparker: Sure. My point is also that at least it is worth a try if try has not yet been tried.

lmuntaner:
Not needed anymore, we're going to use media queries.

PR #331, opened 2023-11-09
Feat: Add new table colors

lmuntaner:
# Motivation

Use CSS variables for the colors of the new table in the Tokens page.

In this PR, add variables in the dark and light themes.

# Changes

* Add four new CSS variables for colors in light and dark themes.

# Screenshots

No changes.

### DIFF
--- a/src/lib/styles/themes/dark.scss
+++ b/src/lib/styles/themes/dark.scss
@@ -12,8 +12,8 @@
   --button-primary: var(--purple-dark-accent);
   --table-background: var(--purple-dark-550);
   --table-header-background: var(--purple-dark-450);
-  --table-row-hover: var(--purple-dark-500);
-  --table-row: var(--purple-dark-550);
+  --table-row-background: var(--purple-dark-550);
+  --table-row-background-hover: var(--purple-dark-500);
 
   // -------- OLD CSS VARIABLES ----------------
   // Shadows
--- a/src/lib/styles/themes/dark.scss
+++ b/src/lib/styles/themes/dark.scss
@@ -10,6 +10,10 @@
   --text-text: var(--purple-900);
   --input-focus-background: var(--purple-dark-450);
   --button-primary: var(--purple-dark-accent);
+  --table-background: var(--purple-dark-550);
+  --table-header-background: var(--purple-dark-450);
+  --table-row-hover: var(--purple-dark-500);
+  --table-row: var(--purple-dark-550);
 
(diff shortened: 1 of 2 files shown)


lmuntaner:
@dskloetd as agreed with Artem, I added the new colors to gix-components.

src/lib/styles/themes/dark.scss
```diff
@@ -10,6 +10,10 @@
   --text-text: var(--purple-900);
   --input-focus-background: var(--purple-dark-450);
   --button-primary: var(--purple-dark-accent);
+  --table-background: var(--purple-dark-550);
+  --table-header-background: var(--purple-dark-450);
+  --table-row-hover: var(--purple-dark-500);
+  --table-row: var(--purple-dark-550);
```
dskloetd, line 16: Are these last 2 also backgrounds? Should we not add `background` to their names as well?

dskloetd, line 16: Optionally you could point this one to the other one like this:
```
--table-row: var(--table-background);
```
But we can also wait with that until we add more layers.

lmuntaner: I'd rather wait to add layers.

And those should probably be set by the designer.

lmuntaner:

src/lib/styles/themes/dark.scss
```diff
@@ -10,6 +10,10 @@
   --text-text: var(--purple-900);
   --input-focus-background: var(--purple-dark-450);
   --button-primary: var(--purple-dark-accent);
+  --table-background: var(--purple-dark-550);
+  --table-header-background: var(--purple-dark-450);
+  --table-row-background: var(--purple-dark-550);
+  --table-row-background-hover: var(--purple-dark-500);
```
dskloetd, line 16: I think for background colors, `background` should be the suffix, so `hover` should come before it.
For example, there is also `--button-card-focus-background` where `focus` comes before `background`.

dskloetd (approved):

src/lib/styles/themes/dark.scss
lmuntaner: I don't have a strong opinion, but I agree that we should have consistent naming conventions.

src/lib/styles/themes/dark.scss
lmuntaner: These are the same name as in Figma, but let me ask Artem to change it. It makes sense to me.

PR #335, opened 2023-11-30
Some more color constants

dskloetd:
# Motivation

1. When I added pending orange colors in https://github.com/dfinity/gix-components/pull/334, I forgot to add them to the dark theme.
2. In NNS-dapp we have an explicit [selector for dark theme](https://github.com/dfinity/nns-dapp/blob/39ce835d00c876698d60df839743f70b98dc3eb9/frontend/src/lib/components/accounts/TransactionCard.svelte#L152C3-L152C28) in the `TransactionCard`. Instead we should use themed color constants. Otherwise the dark theme selector has higher specificity than the `.pending` selector and pending transaction won't be styled properly on the dark them.
3. I need a constant for the background color of [this card component](https://www.figma.com/file/x8dk5axQUf8Bi43uUsqCHc/NNS-Production-%26-Hand-over?type=design&node-id=475-5799&mode=design&t=N3yI2FVqKa2H2zmw-4).

# Changes

1. Move `--pending-color` and `--pending-background` that were added in https://github.com/dfinity/gix-components/pull/334 to the `NEW CSS VARIABLES` section.
2. Add missing `--pending-color` and `--pending-background` to the dark theme.
3. Add `--check` and `--check-tint` to be used for the receive icon as defined [here](https://www.figma.com/file/x8dk5axQUf8Bi43uUsqCHc/NNS-Production-%26-Hand-over?type=design&node-id=475-5516&mode=design&t=N3yI2FVqKa2H2zmw-4).
4. Add `--island-card-background` for the card component mentioned above.

# Screenshots

Used in an NNS branch:

On light theme:
image

On dark theme:
image

### DIFF
--- a/src/lib/styles/themes/dark.scss
+++ b/src/lib/styles/themes/dark.scss
@@ -15,6 +15,12 @@
   --table-row-background: var(--purple-dark-550);
   --table-row-background-hover: var(--purple-dark-500);
 
+  --check: var(--green);
+  --check-tint: var(--green-dark);
+  --pending-color: var(--orange-mid);
+  --pending-background: var(--deep-violet-tint);
+  --island-card-background: var(--purple-dark-550);
+
   // -------- OLD CSS VARIABLES ----------------
   // Shadows
   // TODO-colors: Remove along the dark shadows.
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
Thank you!

PR #342, opened 2023-12-18
Fix redundant toast scrollbar

mstrasinskis:
# Motivation

Get rid of the unnecessary toast scrollbar (screen 1.)

# Changes

- use `line-height: normal;` (presumably the custom font has not default `line-height` which results in an error on browser calculation for `scroll-y: auto`)
- simplify `max-height`, because the line-height is known
- add a couple of toast samples to test the covered cases

# Screenshots

## 1. Before
image

## 2. After

image

## 3. With a title

image


peterpeterparker (approved):
LGTM, thx

PR #343, opened 2023-12-19
Add data-tid to transaction icons

dskloetd:
# Motivation

We want to verify that we render the correct icons for transactions in the nns-dapp.

# Changes

Add `data-tid` attribute to some icons that are used for transaction rendering.

This is not the first time. [Example](https://github.com/dfinity/gix-components/blob/708742f87db44177f9e025dc3df05f241d44eaf9/src/lib/icons/IconStackedLineChart.svelte#L9), used [here](https://github.com/dfinity/nns-dapp/blob/a1598bc03bcebdf8460fcda3ccb67d98a14fd6c9/frontend/src/tests/lib/components/neurons/NnsNeuronCard.spec.ts#L106).

# Screenshots

N/A


mstrasinskis (approved):
Nice, thanks!

PR #344, opened 2023-12-27
style: content-cell-island--highlighted style

mstrasinskis:
# Motivation

Add `content-cell-island__card` style to replace multiple custom styles (ex. `markdown-container`) in `nns-dapp`. 

# Changes

- class `content-cell-island__card`

# Screenshots

image

### DIFF
--- a/src/lib/styles/global/utility-classes.scss
+++ b/src/lib/styles/global/utility-classes.scss
@@ -44,6 +44,11 @@
 
   padding: var(--padding-2x);
   border-radius: var(--border-radius);
+
+  &--highlighted {
+    background: var(--card-background-disabled);
+    color: var(--description-color);
+  }
 }
 
 // Spacing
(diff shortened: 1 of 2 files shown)


src/lib/styles/global/utility-classes.scss
```diff
@@ -44,6 +44,11 @@
 
   padding: var(--padding-2x);
   border-radius: var(--border-radius);
+
+  &--highlighted {
+    background: var(--card-background-disabled);
```
peterpeterparker, line 49: can we interfer another colod? usigin "disabled" for "highlighted" is a bit suprising without context I guess

peterpeterparker (approved):
LGTM, thx

src/lib/styles/global/utility-classes.scss
mstrasinskis: Good point, was replaced w/ `var(--island-card-background);`

peterpeterparker: Cool 👍

PR #345, opened 2023-12-30
feat: menu logo and oneline customizable

peterpeterparker:
# Motivation

Samer reported on the [forum](https://forum.dfinity.org/t/nns-dapp-ui-kit/16298/4?u=peterparker) that it would be handy for they use case to make the logo (and oneliner) of the menu customizable because currently they have to fork Gix components to do so.

# Changes

- new `menu-logo` and `menu-oneliner` on the `Layout` cmp passed down to the `Menu`.
- if no logo is provided, the default logo NNS and oneliner are used.
- remove usage of `HeaderTitle` in the documentation as this PR also display a custom logo in the menu now.


mstrasinskis (approved):
LGTM, thank you!

PR #346, opened 2023-12-30
build: adapt alias for sveltekit v2

peterpeterparker:
# Motivation

Resolve lint warning:

> You have specified a baseUrl and/or paths in your tsconfig.json which interferes with SvelteKit's auto-generated tsconfig.json. Remove it to avoid problems with intellisense. For path aliases, use `kit.alias` instead: https://kit.svelte.dev/docs/configuration#alias

Capture d’écran 2023-12-30 à 09 01 20

### DIFF
--- a/svelte.config.js
+++ b/svelte.config.js
@@ -15,6 +15,9 @@ const config = {
 
   kit: {
     adapter: adapter(),
+    alias: {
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
Thanks!

PR #348, opened 2024-01-08
Design: Color Chages

meodai:
# Changes

Introduce new colors for both dark and light theme. Also inline two of the logos, so they don't need to be there twice.

# Screenshots
Screenshot 2024-01-08 at 14 25 18
Screenshot 2024-01-08 at 14 25 09
Screenshot 2024-01-08 at 14 24 59
Screenshot 2024-01-08 at 14 24 50


meodai:
@peterpeterparker can you make sure the inline logos will work on the actual NNS dapp?

src/lib/components/LogoNNS.svelte
```diff
@@ -1,17 +1,109 @@
-<script lang="ts">
-  import logoNNS from "$lib/assets/logo-nns.svg";
-</script>
-
-<img
+<svg
+  viewBox="0 0 160 34"
+  fill="none"
+  xmlns="http://www.w3.org/2000/svg"
   class="logo-nns"
```
peterpeterparker, line 5: this class can be removed I think

src/lib/components/MenuBackground.svelte
```diff
@@ -80,8 +62,11 @@
   .on-chain {
     // We do not want to display the logo "on chain" behind the menu items
     display: none;
-    padding: 0 0 var(--padding-3x);
+    margin-bottom: var(--padding-3x);
+    width: 60%;
```
peterpeterparker, line 66: not sure if `width` and `opacity` should be defined here because if a developer provide their own "one liner" they will also inherits it.

maybe those style should be applied on the default onliner what do you think?

meodai: Ah yeah of course! I could a) use inline styles or b) have a special class on our own logo. What would you like better?

meodai: oh, but they would not use the on-chain class for their own-liner no? I think we are safe.

peterpeterparker (approved):
LGTM, thanks

PR #349, opened 2024-01-09
build: release v4

peterpeterparker:
# Motivation

Both light and dark mode theme have been reviewed and their color differences is quite predominant therefore we think it is worth releasing those as a major version.

# Changes

- bump version v4 for release


mstrasinskis (approved):
Thank you!

PR #350, opened 2024-01-10
Add QR code placeholder icon

dskloetd:
# Motivation

We want to use this icon instead of a real QR code on the ckBTC wallet page when the user is not logged in.

# Changes

Add `IconQrCode.svelte`.

# Screenshots

Icon:

image

Mockup of how it will be used:
image


src/lib/icons/IconQRCode.svelte
```diff
@@ -0,0 +1,283 @@
+<!-- source: DFINITY foundation -->
+<script lang="ts">
+  // Default to 100% size to make sure Safari renders the SVG correctly
+  export let size = "100%";
```
mstrasinskis, line 4: Will it be used with the default value `100%` as the QR replacement?

dskloetd: I think so.
I put this here to be consistent with `IconAccountsPage` which is a similar kind of icon:
https://github.com/dfinity/gix-components/blob/e04488a9f4de539cd4da2793eb4f282461b0772a/src/lib/icons/IconAccountsPage.svelte#L4

mstrasinskis (approved):

PR #351, opened 2024-01-10
fix(IDX-3073): allow Commit Formatting changes to trigger CI

cgundy:
# Motivation

IDX has been informed that when the CI job `Commit Formatting changes` is triggered, the PR is blocked by the CLA check workflow. This is because the github action `add-and-commit` intentionally does not trigger CI. See explanatation [here](https://github.com/EndBug/add-and-commit?tab=readme-ov-file#the-commit-from-the-action-is-not-triggering-ci):

```
GitHub sees that the push event has been triggered by a commit generated by CI, and doesn't run any further checks to avoid unintentional check loops.
```

However, because the `CLA Check` is required to run before the PR can be merged, the PR is blocked from merging. Previously this issue was resolved by creating an empty commit to re-trigger CI.

# Changes

This change adds a fine-grained Personal Access Token (PAT) to allow the `add-and-commit` action to commit to the PR and trigger CI. This is the solution described in their docs [here](https://github.com/EndBug/add-and-commit?tab=readme-ov-file#the-commit-from-the-action-is-not-triggering-ci).

I have solved this by extending the repo scope of `GIX_CREATE_PR_PAT` to include `gix-components` and granting the `sa-github-api` write access to the repository.

I have also updated the author to be github_actions instead of an empty author.

If you do not want these changes, feel free to close this PR, but you will need to resolve any CLA issues with the empty-commit solution described earlier.

# Screenshots

CI now runs on this commit.
Screenshot 2024-01-11 at 13 19 48

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -36,6 +36,8 @@ jobs:
           default_author: github_actions
           message: "Updating formatting"
           # does not pull by default
+          push: origin HEAD:${{ github.head_ref }}
+
 
   build:
     runs-on: ubuntu-20.04
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -21,6 +21,7 @@ jobs:
           GIX_CREATE_PR_PAT: ${{ secrets.GIX_CREATE_PR_PAT }}
         run: |
           git remote set-url origin https://$GIX_CREATE_PR_PAT@github.com/dfinity/gix-components.git
+          git config --global user.name "sa-github-api"
           git remote show origin
 
       - name: Install shfmt
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -9,6 +9,8 @@ on:
 jobs:
   formatting:
     runs-on: ubuntu-20.04
+    permissions:
+      contents: write
 
     steps:
       - name: Checkout
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -20,7 +20,7 @@ jobs:
         env:
           GIX_CREATE_PR_PAT: ${{ secrets.GIX_CREATE_PR_PAT }}
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/checks.yml
```diff
@@ -26,18 +28,18 @@ jobs:
       - name: Format
         run: npm run format
 
-      - name: Commit Formatting changes - will not trigger rebuild
+      - name: Commit Formatting changes
```
cgundy, line 31: I am not sure what was meant by the comment "will not trigger rebuild" but this action will now trigger CI.

peterpeterparker (approved):
That's neat, thanks!

We probably got the same issue in https://github.com/dfinity/ic-js given that the actions were copied from there. 

Moreover, the actions in ic-js were copied from nns-dapp. Maybe there is the issue there as well.

Can you also check those two other projects?

cgundy:
Sure, I'll take a look 👍

.github/workflows/checks.yml
```diff
@@ -26,18 +28,18 @@ jobs:
...
           add: .
-          author_name: Formatting Committer
-          author_email: "<nobody@example.com>"
+          default_author: github_actions
           message: "Updating formatting"
           # do not pull: if this branch is behind, then we might as well let
           # the pushing fail
           pull_strategy: "NO-PULL"
 
+
```
dskloetd, line 42: Do we need a double blank line here?

peterpeterparker:
Thanks!

dskloetd (approved):
Thanks a lot!
Sorry, I got distracted while reviewing. Probably not worth making another PR but I left a comment already before I saw it was merged.

.github/workflows/checks.yml
cgundy: Ah sorry, that was an accident.

cgundy:
@peterpeterparker looks like `ic-js` has the same structure and I updated it, but `nns-dapp` has a different CI that doesn't use the `add-and-commit` action. But feel free to ping IDX if you see issues there.

PR #352, opened 2024-01-12
Feat: Remove z-index from button

lmuntaner:
# Motivation

There is an issue with the content of the button going over another element.

Screenshot 2024-01-12 at 11 43 17

The problem is the `z-index` set in the `ghost` and `text` button utilities.

# Changes

* Remove the `z-index` from the button styles.

# Screenshots

No UI changes.

### DIFF
--- a/src/lib/styles/global/button.scss
+++ b/src/lib/styles/global/button.scss
@@ -9,8 +9,6 @@ button {
 
   &.ghost,
   &.text {
-    z-index: var(--z-index);
-
     width: fit-content;
 
     @include fonts.standard;


lmuntaner:
@dskloetd @peterpeterparker please review

@peterpeterparker was there a reason for this z-index?

I have tested this branch with NNS Dapp checking where the ghost button is used and it all seems good.

peterpeterparker:
> was there a reason for this z-index?

I would bet they appeared over some backdrop for spinner or modals or weren't correctly indexed in the menu, something like that. But if it's fine now, super.

PR #357, opened 2024-01-17
fix: trim currency decimals as string

peterpeterparker:
# Motivation

While testing the new input type `currency` in Oisy I landed on an that was parsed to number with scientific notation instead of binding the value to string. This happened when I entered a number with the max number of decimals and then trying to enter additional decimals.

For example: I entered 0.000000011231232121 in the input field, 18 decimals, and then tried to add a 19th decimals.

# Changes

- in the input field hook `restoreFromValidValue`, map to number only for the `inputType` is `icp` - i.e. if `currency` keep string
- use renamed `decimals` property in the showcase

# Screenshots

Capture d’écran 2024-01-17 à 07 43 26

### DIFF
--- a/src/lib/components/Input.svelte
+++ b/src/lib/components/Input.svelte
@@ -91,7 +91,9 @@
       ? undefined
       : typeof lastValidCurrencyValue === "number"
       ? lastValidCurrencyValue.toFixed(wrapDecimals)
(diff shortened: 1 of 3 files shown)


mstrasinskis (approved):
Thanks!

PR #359, opened 2024-01-19
Add test ID to Copy component

dskloetd:
# Motivation

To test if the component is present in the DOM.

# Changes

Added `data-tid` attribute to `Copy` component.

# Screenshots

N/A


lmuntaner (approved):
Thanks

PR #361, opened 2024-01-22
fix: input currency decimals rounding and trim zero

peterpeterparker:
# Motivation

While testing the new `currency` mode of the `Input` property in Oisy, I noticed two issues which are corrected by this PR.

Leading zeros were added to the binded property (entering 0. lead to 0.000000000000) and JS was rounding the number imprecisely even though this mode requires a string as output.

# Changes

- use internationalization to wrap the decimals of the "currency"

# Screenshots

Rounding issue:
Capture d’écran 2024-01-22 à 08 08 44

Trim issue:
Capture d’écran 2024-01-22 à 08 08 40

Fix:
Capture d’écran 2024-01-22 à 07 57 00

### DIFF
--- a/e2e/input.e2e.ts
+++ b/e2e/input.e2e.ts
@@ -1,0 +1,13 @@
+import { expect, test } from "@playwright/test";
+
+const testUrl = "/components/input";
+
+test("should trim zero", async ({ page }) => {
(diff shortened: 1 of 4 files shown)


mstrasinskis (approved):
Nice, thank you!

PR #362, opened 2024-01-22
Feat: Add class utility to link to look like buttons

lmuntaner:
# Motivation

Add a utility class for the anchor tags to look like a primary button.

This is used in the Buy ICP flow of the NNS Dapp. It is also used with an icon. Therefore, I also added the `with-icon` utility class to both the link and button utility classes.

# Changes

* New class utility for anchor tag: `button` + `primary` + `full-width` + `with-icon`.
* Add `with-icon` to the button utility classes.
* LInk page docs.
* Add a card in Utilities linking to the link docs.
* Add e2e screenshot test for the links docs.
* Add docs about `with-icon` in the button docs.

# Screenshots

Check the screenshots in the PR.


lmuntaner:
@peterpeterparker @dskloetd please review

src/routes/(page)/utility-classes/links/+page.md
```diff
@@ -0,0 +1,30 @@
...
+## Showcase
+
+### Default
+
+<div class="container" data-tid="showcase">
+
+| Styles                                                                            |
+| --------------------------------------------------------------------------------- |
+| <a href="#">Default Link</a>                                                      |
+| <button class="primary">Primary Button</button>                                   |
```
dskloetd, line 19: Why is there a button here?

dskloetd (approved):

src/routes/(page)/utility-classes/links/+page.md
lmuntaner: I added it for users to be able to check that it's indeed the same style. But true that it's not necessary...

src/lib/styles/global/link.scss
```diff
@@ -13,4 +13,48 @@ a {
   &:hover {
     color: var(--value-color);
   }
+
+  &.button {
```
mstrasinskis, line 17: Just thinking out loud, would it make sense to create a button `mixin` to share some paddings and colors?

lmuntaner: Yes, I agree. Let me do it.

mstrasinskis: Thank you!

src/lib/styles/global/button.scss
```diff
@@ -1,12 +1,17 @@
 @use "../mixins/effect";
 @use "../mixins/fonts";
+@use "../mixins/button";
```
meodai, line 3: i'd rather to a class called .button that can be set on a link than a mixin.

src/lib/styles/global/link.scss
```diff
@@ -13,4 +14,35 @@ a {
...
```
meodai, line 18: why should the class .button only work on links? seams overly specific.

src/lib/styles/global/button.scss
lmuntaner: That's a good point.

I believe that supporting all the other button types would be a bit more work than what I was expecting in this PR.

meodai: But like this you are actively preventing from it being used anywhere else + the code is there twice, once in the button css and one in tha a.button css.

src/lib/styles/global/link.scss
lmuntaner: done!

mstrasinskis (approved):
Thanks!

PR #363, opened 2024-01-23
replace color from notion so it would work in firefox

meodai:
# Motivation

the CSS rgb(`from <something>`) is not working in Firefox yet

# Changes

Use the more traditional hex notation to be constant with the other colors.

# Screenshots
Screenshot 2024-01-23 at 11 26 53


lmuntaner (approved):
Thanks!

PR #365, opened 2024-01-23
build: fetch QRCode test video from dfinity download

peterpeterparker:
# Motivation

Previous Google Drive hosting is for some reason broken and download required some logic anyway so we moved the file to dfinity download.

Also it looks like last PR was merged with failing test. Not sure if it was a glitch or recent changes in the pipeline would lead to this. We should keep an eye on it. As a results, this PR also update all required screenshots.

# Notes

Fee overlays background in light mode appears to have gone darker in light mode. I pinged David about it. Should not be a blocker for this PR which has for goal to fix the CI.


lmuntaner (approved):
Only four seemed to have failed in the last PR: https://github.com/dfinity/gix-components/actions/runs/7624340823/job/20766271507?pr=363

Why would all change?

peterpeterparker:
> Why would all change?

Local (Mac, darwin) screenshots vs GitHub Actions (linux). I guess they weren't regenerated in previous PR(s).

PR #367, opened 2024-01-25
Copy Tooltip from nns-dapp

dskloetd:
# Motivation

NNS dapp has a tooltip component with somewhat broken behavior.
Before fixing it, I thought it would be useful to have a showcase in gix-components to easily play around with it while changing it.
And it seemed like the kind of component that would be useful to have in gix-components anyway.

So before fixing the tooltip component, I'm copying it unchanged from nns-dapp to gix-components.

Once this component is fixed, we can switch nns-dapp over to this tooltip component.

# Changes

1. Copied `Tooltip.svelte`, `Tooltip.spec.ts` and `TooltipTest.svelte` from the `nns-dapp` repo unchanged.
2. Added a page with showcase for the component.

# Screenshots

image


lmuntaner (commented):
It would be good to add an e2e test, which are easy to add here.

dskloetd:
> It would be good to add an e2e test, which are easy to add here.

Done

lmuntaner (approved):
We could also have a screenshot once we fix the issue, right?

dskloetd:
> We could also have a screenshot once we fix the issue, right?

What do you mean? Did you not see the screenshot?

lmuntaner:
> > We could also have a screenshot once we fix the issue, right?
> 
> What do you mean? Did you not see the screenshot?

Sorry, I meant a screenshot testing of the issue once it's fixed.

dskloetd:
Actually, the e2e test I added observes the issue.
The screenshot when taken on Mac looks like the screenshot in the PR description.
But on Linux it somehow doesn't trigger.
My guess is that the button is right in the center and through some kind of rounding error is considered to be left of center on Linux and right of center on Mac or something.

But once it's fixed, the issue of course won't be visible in the screenshot because then the issue doesn't exist anymore.
So I'm still not sure what kind of screenshot you expect after fixing the isuse.

lmuntaner:
> Actually, the e2e test I added observes the issue. The screenshot when taken on Mac looks like the screenshot in the PR description. But on Linux it somehow doesn't trigger. My guess is that the button is right in the center and through some kind of rounding error is considered to be left of center on Linux and right of center on Mac or something.
> 
> But once it's fixed, the issue of course won't be visible in the screenshot because then the issue doesn't exist anymore. So I'm still not sure what kind of screenshot you expect after fixing the isuse.

I would expect a working screenshot that in case we break it again we find it early.

lmuntaner:
> Oh, I see. I don't think that's what screenshots are for. And also it's more likely that some cases we didn't think of are still/already broken rather than that some obvious case break. And it would be too many screenshots to cover even a decent fraction of cases.
> 
> The right way forward is to factor the logic outside of the component and have lots of unit tests with numbers.

If we can make it as unit tests better, yes.

PR #373, opened 2024-01-30
Button class text for links

mstrasinskis:
# Motivation

Add `a.button.text` class to mimic text buttons for links.

# Changes

- moved `button.text` to mixins to reuse for links
- add `a.button.text`

# Screenshots

image

### DIFF
--- a/src/lib/styles/global/link.scss
+++ b/src/lib/styles/global/link.scss
@@ -16,7 +16,13 @@ a {
   }
 
   &.button {
-    @include button.base;
+    &.primary,
+    &.secondary,
+    &.success,
+    &.danger,
+    &.warning {
+      @include button.base;
+    }
 
     // Default is `inline` and buttons are `inline-block` but they have a hidden flex implemented in the browser.
     // https://stackoverflow.com/questions/15764600/what-makes-the-text-on-a-button-element-vertically-centered/34612505#34612505
@@ -28,6 +34,10 @@ a {
     box-sizing: border-box;
 
     text-decoration: none;
+
+    &.text {
+      @include button.text;
+    }
 
     &.primary {
       @include button.primary;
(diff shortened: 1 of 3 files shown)


src/lib/styles/global/link.scss
```diff
@@ -16,7 +16,13 @@ a {
   }
 
   &.button {
-    @include button.base;
+    &.primary,
+    &.secondary,
+    &.success,
+    &.danger,
+    &.warning {
+      @include button.base;
```
mstrasinskis, line 24: Same as for button to avoid extra paddings on `a.button.text`

src/lib/styles/mixins/_button.scss
```diff
@@ -33,6 +35,27 @@
   }
 }
 
+@mixin text {
+  width: fit-content;
+
+  @include fonts.standard;
+
+  text-decoration: none;
```
mstrasinskis, line 43: Set default `text-decoration` value to avoid links style.

src/lib/styles/global/link.scss
mstrasinskis: Looks like currently there is only one place where it's used, so not a breaking change `<a class="button primary full-width with-icon"...`.

src/lib/styles/global/link.scss
```diff
@@ -16,7 +16,13 @@ a {
   }
 
   &.button {
-    @include button.base;
+    &.primary,
```
lmuntaner, line 19: Why not using `.button:not(.text)` instead?

lmuntaner (approved):
Only one comment, LGTM. Thanks!

src/lib/styles/global/link.scss
mstrasinskis: To follow the same structure as in `buttons.scss` [@include button.base;](https://github.com/dfinity/gix-components/blob/f5dcbb37797595af47afde13e15476670e0e6902/src/lib/styles/global/button.scss#L61)

lmuntaner: It makes sense to be consistent. Thanks!

PR #374, opened 2024-01-30
style: increase popover max-width

mstrasinskis:
# Motivation

Increase `Popover` max-width to support wider content (like a redesigned account menu - screenshot)

# Changes

- replace fixed width with all `available space - right` to make it nicely shrink on mobile (screen 2).

# Screenshots

## Enough space
image

## Not enough space
image

### DIFF
--- a/src/lib/components/Popover.svelte
+++ b/src/lib/components/Popover.svelte
@@ -10,7 +10,6 @@
   export let visible = false;
   export let direction: "ltr" | "rtl" = "ltr";
   export let closeButton = false;
-  export let noMaxWidth = false;
 
   let bottom: number;
   let left: number;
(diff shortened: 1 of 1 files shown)


src/lib/components/Popover.svelte
```diff
@@ -98,6 +100,11 @@
     border-radius: var(--border-radius);
 
     box-shadow: var(--overlay-box-shadow);
+
+    &.noMaxWidth {
+      // 100vw - right padding
+      max-width: calc(100vw - var(--padding));
```
peterpeterparker, line 106: Do we actually need `max-width: var(--size);` and if yes, do you remember why?

If we don't need it, maybe it should always be `max-width: calc(100vw - var(--padding));`?

I generally like to avoid JS to apply a UI behavior on a single CSS element, I would rather user a CSS variable, so just try to understand if there is not another way.

mstrasinskis: Because of [this popover](https://github.com/dfinity/nns-dapp/pull/2764) the fixed max-width should be still supported.
image

peterpeterparker: Thanks for the feedback

peterpeterparker (approved):
LGTM, discussed offline. Thanks.

PR #375, opened 2024-01-31
Toolbar buttons

mstrasinskis:
# Motivation

A mixin to solve the button with tooltip alignment issue.

# Changes

- new mixin `toolbar`.
- replace custom css in modals with the `toolbar` mixin.

# Screenshots

Disburse maturity. The jumping is another issue because of the tooltip position.
Screen Recording 2024-01-31 at 12 02 05

Participation button
image


src/lib/styles/global/modal.scss
```diff
@@ -14,21 +14,8 @@ div.modal {
   }
 
   .toolbar {
-    display: flex;
```
dskloetd, line 17: Does this mean that existing usages of `.toolbar` will break unless they also add `.toolbar-buttons`?

mstrasinskis: Yes, the plan was to add the `.toolbar-buttons` class together with the `gix-upgrade` pr.

dskloetd: What do you think about moving the `toolbar` class outside of `div.modal` so we can reuse it without having to add multiple classes each time?
Or would that not work?

mstrasinskis: Not sure, 
1. it has an extra padding inside the modal;
2. it's a bit confusing that there will be a global class `.toolbar` that is not related directly to the `Toolbar` component.

What do you think about renaming `.toolbar-buttons` in this pr to `.full-width-mobile-buttons`?

dskloetd: I'm concerned that using `.toolbar` inside `div.modal` used to work and now requires an extra class.

How about also adding `.toolbar` style that works inside `BottomSheet`?

I'm also still confused how this even works. Why does the `div.modal` selector here work without the need for `:global`?

dskloetd: Then why did it have `:global(button) {` on line 22?

mstrasinskis: Good question, but it doesn't work w/o `:global`

dskloetd: Thanks for checking.

Are you still looking into this:
> I'm concerned that using .toolbar inside div.modal used to work and now requires an extra class.
> How about also adding .toolbar style that works inside BottomSheet?

Or did you miss that?

mstrasinskis: A [mixin](https://github.com/dfinity/gix-components/pull/375/files#diff-bbba86cbcfd780d497c1fc89e609c3599beee6594f8bdb01bcc8195210e97565R20) was added.
- no changes needed for modals
- can be used to fix the "Participate" button issue -- https://github.com/dfinity/nns-dapp/pull/4401

mstrasinskis: @dskloetd, pr was updated:
- `*` selectors to make it most flexible
- keep only `base` version of the mixin

src/lib/styles/mixins/_toolbar.scss
```diff
@@ -0,0 +1,28 @@
+@use "./media";
+
+// A mixin to align buttons (also wrapped).
+// Uses all available space to aligns the buttons on mobile. Don't resize/alight buttons on desktop.
```
dskloetd, line 4: typo: alight

dskloetd (approved):
Thanks!

src/lib/styles/mixins/_toolbar.scss
mstrasinskis: Thanks.

PR #376, opened 2024-02-05
feat: browser detection

peterpeterparker:
# Motivation

A page that can be used to determine capatibilities of the browsers used by users of NNS dapp / gix-cmp.

Useful for example to debug issue related to device detection such as if a device is mobile or desktop.

Demo: https://gix.design/browser


mstrasinskis (approved):
Thanks!

PR #377, opened 2024-02-05
feat: ignore 1Password in input fields

peterpeterparker:
# Motivation

Disable per default 1Password in input fields.

# Changes

- add ignore flag as described in 1Password documentation https://developer.1password.com/docs/web/compatible-website-design/

# Screenshots

This solve following:

Capture d’écran 2024-02-05 à 10 28 19

### DIFF
--- a/src/lib/components/Input.svelte
+++ b/src/lib/components/Input.svelte
@@ -14,6 +14,7 @@
   export let placeholder: string;
   export let testId: string | undefined = undefined;
   export let decimals = 8;
+  export let ignore1Password = true;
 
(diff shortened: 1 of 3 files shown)


peterpeterparker:
In Oisy feels correct to disabled the plugin per default, see screenshot. If you think we should rather do the contrary let me know.

peterpeterparker:
In Oisy it's correct to disabled the 1pwd plugin per default, that's why this PR do so but, if you think we should rather do the contrary let me know @mstrasinskis.

mstrasinskis (approved):
Thanks!

PR #378, opened 2024-02-06
feat: improve mobile and desktop detection for QRCode reader mirroring

peterpeterparker:
# Motivation

We received reports from users complaining about the QRCode reader being incorrectly mirrored on their devices. This issue was specifically noted on Samsung devices that support a stylus. The presence of a stylus leads the device to report to the user agent that it supports `any-pointer:fine`, i.e. it indicates support for a "mouse", which we previously used to detect desktop environments.

This PR revises our method for detecting mobile devices by utilizing the `NavigatorUAData` agent, which is available in Chrome. It also updates our approach for identifying desktop devices to correct the QRCode reader mirroring issue.

# Tradeoff

This implementation comes with a tradeoff. Since `NavigatorUAData` is not yet supported in Firefox, using this browser on a touchscreen device will result in it being detected as a mobile device.

This might be considered an edge case, given the current state of the APIs, this implementation represents our most accurate method for device detection available at this time.

# Resources

- https://caniuse.com/mdn-api_navigator_useragentdata
- https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData
- https://stackoverflow.com/questions/11381673/detecting-a-mobile-browser

# Changes

- mirror qrcode reader if `isDesktop`
- review `isMobile`
- add TS support for experimental feature

### DIFF
--- a/src/app.d.ts
+++ b/src/app.d.ts
@@ -1,4 +1,5 @@
 /// <reference types="@sveltejs/kit" />
+/// <reference types="user-agent-data-types" />
 
 // See https://kit.svelte.dev/docs/types#app
 // for information about these interfaces
--- a/src/lib/utils/device.utils.ts
+++ b/src/lib/utils/device.utils.ts
@@ -1,4 +1,5 @@
 import { isNode } from "$lib/utils/env.utils";
+import { nonNullish } from "@dfinity/utils";
 
 export const isPortrait = (): boolean => {
   if (isNode()) {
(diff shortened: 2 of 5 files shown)


src/app.d.ts
```diff
@@ -1,4 +1,5 @@
 /// <reference types="@sveltejs/kit" />
+/// <reference types="user-agent-data-types" />
```
mstrasinskis, line 2: Out of curiosity, have you added this line manually?

src/lib/utils/device.utils.ts
```diff
@@ -56,18 +57,27 @@ export const isAndroidTablet = (): boolean => {
...
-  const isMouseScreen: boolean =
-    window.matchMedia("(any-pointer:fine)").matches;
+  // Available in Chrome on any devices (Feb 2024)
+  // https://caniuse.com/mdn-api_navigator_useragentdata
+  if ("userAgentData" in navigator && nonNullish(navigator.userAgentData)) {
+    return navigator.userAgentData.mobile;
+  }
+
+  // Legacy
+  const isTouchScreen = window.matchMedia("(any-pointer:coarse)").matches;
```
mstrasinskis, line 76: Let's hope there will be no macbooks w/ a touchscreen :D

mstrasinskis (approved):
Thank you, let's try!

src/app.d.ts
peterpeterparker: My editor wasn't resolving the types and making the types visible is a step of the installation guide: https://github.com/lukewarlow/user-agent-data-types#make-types-visible-globally-in-all-source-files-within-a-project

src/lib/utils/device.utils.ts
peterpeterparker: Indeed!

Worth to note that this "legacy" approach is similar to (not legacy) approach of [Ionic](https://github.com/ionic-team/ionic-framework/blob/35ab6b4816bd627239de8d8b25ce0c86db8c74b4/core/src/utils/platform.ts#L92).

mstrasinskis: Ah, a good reference, thanks!

PR #379, opened 2024-02-06
Feat: Support html in toasts

lmuntaner:
# Motivation

Support html strings in toast messages. Specially useful for links to documentation.

# Changes

* Use `Html` for the `text` inside a `Toast`.
* Add an example in the docs page.

# Screenshots

Screenshot 2024-02-06 at 13 21 13


lmuntaner:
@dskloetd @mstrasinskis please review

src/lib/components/Toast.svelte
```diff
@@ -102,7 +103,7 @@
     {#if nonNullish(title)}
       <span class="title">{title}</span>
     {/if}
-    {text}
+    <Html {text} />
```
dskloetd, line 106: I wouldn't do this if you don't know what `text` is.
Can we use a different field in case we want to use HTML?
Not only might we have toasts with user input, this gix-components might be used by other applications that we don't even know.
I know the HTML is validated so probably it won't cause literal XSS but even just messing with the UI or just not rendering `<` and `>` properly is not good.

lmuntaner: The `<` and `>` are rendered if alone:

Screenshot 2024-02-06 at 15 18 44

What do you suggest then? Another property in the message type?

dskloetd: I'm not sure what is the message type you're referring to, but something like that, yes.
When creating such a toast, you should either use a different function or a different argument to the function. If you want HTML you should be explicit about it.

lmuntaner: This is the [Message type](https://github.com/dfinity/gix-components/blob/4092ec5ba6ecf2b604d86c3c0c03cdbd7158088f/src/lib/types/toast.ts#L9) I meant.

If we have two fields, shall we render both?

`text` is mandatory now. How about adding `htmlText` and making both optional?

What are you worried about to render sanitized HTML? Is it because it's an open source library?

dskloetd: 2 optional fields is fine. In that case I would throw an Error if none or both fields are defined.
Another option is to add a boolean `renderAsHtml` which would apply to the existing `text` fields.
I don't have a preference so you can pick what's easier or what you like more.

I'm not worried about it being open source. I'm worried that we promised (to others and ourselves) that something would be rendered as text and we shouldn't break that promise for 2 reasons:
1. If someone wanted to render "Please use <b></b> tags to make text bold.", then this is no longer possible.
2. If we render user input, even with sanitized HTML, the user might still be able to do harm by being able to manipulate the UI. I can't give an example because I don't know what exactly gets sanitized.

lmuntaner: Done, I went for the new optional field `renderAsHtml`. I thought that would be the easiest transition.

lmuntaner:

src/routes/(split)/components/toasts/+page.md
```diff
@@ -63,6 +63,12 @@
         level: "info",
         duration: 3000
     });
+
+    const withHtml = () => toastsStore.show({
```
dskloetd, line 67: How about adding an HTML tag in the multiline message above to demonstrate that it's rendered as text?

And are there no unit tests?

dskloetd (approved):
Thanks!

src/routes/(split)/components/toasts/+page.md
lmuntaner: I thought there were no tests here, my bad.

I preferred to add another button with html rendered as text. So that when reading the button it indicates simply what to expect and do only one thing.

dskloetd (approved):

PR #381, opened 2024-02-06
style: logo spacing

mstrasinskis:
# Motivation

Update nns logo style to make it bigger.

# Changes

- remove svg width to make the logo size based on available space
- update paddings to align to the design

# Screenshots

image

image


src/lib/components/LogoNNS.svelte
```diff
@@ -1,126 +1,127 @@
...
-          width="41.8421"
-          height="19.7368"
-          fill="white"
-          transform="translate(37.8572)"
-        />
-      </clipPath>
-    </defs>
-  </svg>
-</div>
+<svg
```
mstrasinskis, line 1: Same `svg`, only the size attributes have been removed.

src/lib/components/LogoNNS.svelte
```diff
@@ -1,126 +1,127 @@
...
+</svg>
 
 <style lang="scss">
-  div {
-    display: flex;
-    justify-content: center;
+  @use "../styles/mixins/media";
+
+  svg {
+    padding: 0 var(--padding-6x) 0;
```
dskloetd, line 121: I don't think this belongs in the image. Can we add rather the paddings where the image is used?

mstrasinskis: Done.

dskloetd (approved):

PR #382, opened 2024-02-08
style: apply extra styles to the buttons/tooltip only

mstrasinskis:
# Motivation

Update tooltip styles to fix multiple alignment issues:

- set width for all `button` tags `00% on  mobile and control 1 level elements width w/ flex
- reset width for bigger breakpoints to use buttons content size.

# Changes

- styles of .toolbar

# Screenshots

image

---

image

---

image

---

image

---

image

### DIFF
--- a/src/lib/styles/mixins/_toolbar.scss
+++ b/src/lib/styles/mixins/_toolbar.scss
@@ -7,17 +7,21 @@
   display: flex;
   gap: var(--padding-2x);
 
-  & > :global(*) {
+  & > :global(button),
+  & > :global(.tooltip-wrapper) {
     // use 50% to avoid non even button sizes (a single button would be 100%)
     flex: 1 1 50%;
+  }
 
+  & > :global(.tooltip-wrapper) {
     :global(*) {
       width: 100%;
     }
   }
 
   @include media.min-width(small) {
-    & > :global(*) {
+    & > :global(button),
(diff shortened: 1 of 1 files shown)


src/lib/styles/mixins/_toolbar.scss
```diff
@@ -7,23 +7,26 @@
...
+    width: 100%;
   }
 
-  & > :global(.tooltip-wrapper) {
-    :global(*) {
-      width: 100%;
-    }
+  & > :global(*) {
+    // 50% to make sure the button is not smaller in the tooltip, because of the wrong width calculation.
+    flex: 1 1 50%;
```
dskloetd, line 17: Does `50%` assume there are 2 buttons? What if there are 1 or 3?

dskloetd (approved):
If the current code works, this is a fine change.
Just a question about something that was there already.

src/lib/styles/mixins/_toolbar.scss
mstrasinskis: It looks best w/ 1 or 2 buttons, but also not so bad w/ 3+ because of `flex: 1 1` part:

image

image

image

dskloetd: Then I'm not sure why it's `50%` and not something else. Would be good to know but not important for this PR.

PR #386, opened 2024-02-19
Hide tooltip when there is no text

dskloetd:
# Motivation

Sometimes we want to have a tooltip conditional on some logic, such as [here](https://github.com/dfinity/nns-dapp/blob/9f0bd72c6e315adf7443bed1ee14571e600cd2a7/frontend/src/lib/components/neuron-detail/NeuronSelectPercentage.svelte#L65).
It would be convenient if the tooltip was just hidden if there is no text.

# Changes

1. Allow `text` to be undefined and hide the tooltip in that case.
2. Reposition the tooltip if the `text` changes.
3. Change the showcase to allow changing the `text`.
4. Add unit tests.

# Screenshots

Looks the same when it's not hidden.


src/routes/(split)/components/tooltip/+page.md
```diff
@@ -1,19 +1,25 @@
 <script lang="ts">
   import Tooltip from "$lib/components/Tooltip.svelte";
   import Checkbox from "$lib/components/Checkbox.svelte";
+  import Input from "$lib/components/Input.svelte";
 
   let overflowHidden = true;
   let showShowcase = true;
+  let tooltipText = "This button is disabled because of a long and complicated explanation.";
```
lmuntaner: Why did you change the text?

src/tests/lib/components/Tooltip.spec.ts
```diff
@@ -18,4 +21,17 @@ describe("Tooltip", () => {
     ).toBeInTheDocument();
     expect(container.querySelector("[id='tid']")).toBeInTheDocument();
   });
+
+  it("should hide without text", () => {
```
lmuntaner, line 25: Should we also test with `undefined`?

src/tests/lib/components/TooltipTest.svelte
```diff
@@ -1,7 +1,9 @@
 <script>
   import Tooltip from "$lib/components/Tooltip.svelte";
+
+  export let text = "text";
```
lmuntaner, line 4: Shouldn't this be the same type as in the tooltip to allow for `undefined`?

lmuntaner (commented):
A couple of comments

src/routes/(split)/components/tooltip/+page.md
dskloetd: That was an accident that I only realized when updating the screenshot.

src/tests/lib/components/Tooltip.spec.ts
dskloetd: Done

src/tests/lib/components/TooltipTest.svelte
dskloetd: Done

lmuntaner (approved):
LGTM!

src/lib/components/Tooltip.svelte
```diff
@@ -29,6 +29,10 @@
       return;
     }
 
+    if (isNullish(text) || text.length === 0) {
```
mstrasinskis, line 32: Not really important, but this could be a nice local function like `const notText  = () => isNullish(text) || text.length === 0;` Or even a variable.

peterpeterparker: We have a utils `notEmptyString`.

dskloetd: Thanks. Done.

PR #387, opened 2024-02-26
Radio toggle component

mstrasinskis:
# Motivation

A radio group component to be used first on the proposals page.

# Changes

- `RadioToggle` compoent

# Test

- `RadioToggle`

# Screenshots

image

image

### DIFF
--- a/src/docs/constants/docs.constants.ts
+++ b/src/docs/constants/docs.constants.ts
@@ -267,6 +267,12 @@ export const COMPONENT_ROUTES: ComponentRoute[] = [
   },
 
   {
+    path: "/components/radio-toggle",
+    title: "RadioToggle",
+    description: "Radio button group that looks like a toggle.",
+  },
+
+  {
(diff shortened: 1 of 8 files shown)


dskloetd:
Can we perhaps use `Segment` and `SegmentButton`, like we use [here](https://github.com/dfinity/nns-dapp/blob/000bb175142aa1f095a886f0e545c4178d2a6919/frontend/src/lib/modals/accounts/CkBTCReceiveModal.svelte#L130) to switch between ckBTC and Bitcoin on the receive modal?

image

dskloetd:
I think we should be careful about adding new design elements every time we add a new feature and try to reuse components as much as possible. Would be could if you could check with PM/UX if this can be reused. Maybe by adjusting either the existing component or the new feature. Thanks!

mstrasinskis:
Closing this PR as it’s no longer relevant.

PR #388, opened 2024-02-29
Restyle segment component

mstrasinskis:
# Motivation

Restyle the segment component and prepare for usage on proposals page.

# Changes

- update styles
- disable selected button
- [fix the issue when indicator initially always point to the first element](https://github.com/dfinity/gix-components/pull/388/files#diff-72e99015f541e8d0c0016d3a27d888a217d9a63dfa16cfef40128da08b2e8955R24)
- because the component uses full with on mobile, [update position onResize](https://github.com/dfinity/gix-components/pull/388/files#diff-2212548ca2a70ca70d0425f28f85ab2e91ef851bf1c0db9e0dc3fa740359bb55R67) is needed

# Screenshots

## Mobile
image

## Tablet
image

## Desktop
image

## Receive modal

image


src/lib/components/Segment.svelte
```diff
@@ -114,10 +129,19 @@
       -1px,
       0
     );
-    width: calc((100% - (var(--segments) * var(--padding))) / var(--segments));
-    padding: var(--padding-2x) 0 var(--padding);
+    width: calc(
+      (
+        (
+            100% -
```
dskloetd, line 135: This indentation here looks quite random?

src/lib/components/Segment.svelte
```diff
@@ -114,10 +129,19 @@
       -1px,
       0
     );
-    width: calc((100% - (var(--segments) * var(--padding))) / var(--segments));
-    padding: var(--padding-2x) 0 var(--padding);
+    width: calc(
```
dskloetd, line 132: This is quite some logic to put in CSS. Do you want to add a comment explaining what this is doing?

src/lib/components/SegmentButton.svelte
```diff
@@ -17,32 +18,45 @@
 
   let selected = false;
   $: selected = $store.id === segmentId;
+
+  // Update the store with initially selected element as soon as it is available
```
dskloetd, line 22: The code below updates continuously. Should it only update once on init? That's what the comment seems to say.

mstrasinskis: Sorry, I didn't get this. For me this block updates the store only once.

src/lib/components/Segment.svelte
mstrasinskis: Good point, I hope w/ the comment it's better. Just tested it adding extra borders to see sizes:
image

src/lib/components/Segment.svelte
mstrasinskis: Extra parenthesises were removed 👍

src/lib/components/SegmentButton.svelte
dskloetd: This code in a `$:` block is re-evaluated every time any of the values changes.
So every time `selected` or `$store` changes, if the condition holds, it will set the store value. Is this intended?
Or are you saying that these values never change after they get their initial value?

src/lib/components/Segment.svelte
```diff
@@ -114,10 +129,20 @@
       -1px,
       0
     );
-    width: calc((100% - (var(--segments) * var(--padding))) / var(--segments));
-    padding: var(--padding-2x) 0 var(--padding);
+
+    // Calculate the width of the indicator:
+    // (100% - [left and right segment paddings] + [all gaps between buttons]) / [number of segments]
+    width: calc(
+      (
```
dskloetd, line 136: I still don't understand what's going on with the indentation here.
I've added the current indentation in front of every line here:
```
 6      (
10          100% -
12            (
14              2 * var(--segment-padding) + var(--segment-gap) *
16                (var(--segments) - 1)
12            )
 8        ) / var(--segments)
```
Should the parenthesis with 6 indentation be aligned with the one with 8 indentation?

Also, I think the line breaks could be more intuitive to indicate the priority of addition vs. multiplication.
Would this work?
```
      (
        100% -
          2 * var(--segment-padding) +
          var(--segment-gap) * (var(--segments) - 1)
      ) / var(--segments)
```

src/lib/components/SegmentButton.svelte
mstrasinskis: Ah, got your point. The issue this solves is that initially there is no `element` in the store https://github.com/dfinity/gix-components/blob/6952fff6a7099cdd367a850a6bee0881136cc8e5/src/lib/components/Segment.svelte#L12.
But it's already there after clicking on a button, so every time the condition is re-evaluated `$store.element !== element` is false.

mstrasinskis: But we need an element in the store to be able to set the right indicator position. This was a bug of the component before the redesign. Now we need this functionality.

dskloetd (approved):
Thanks!

PR #391, opened 2024-03-01
Colour adjustment 2024 03 01

mstrasinskis:
# Motivation

Colours update.

# Changes

- modal content color (related to the light theme)
- copy icon 
- info icon (in the `KeyValue` component)

# Screenshots

## The modal content block is slightly darker in light theme and looks the same in dark:
| Dark | Light |
|--------|--------|
| image | image | 

## New colours for info and copy icons
| Dark | Light |
|--------|--------|
| image | image |


peterpeterparker (approved):
LGTM, thx

src/lib/styles/themes/dark.scss
```diff
@@ -112,7 +112,7 @@
 
   --overlay-background: var(--background);
   --overlay-background-contrast: var(--background-contrast);
-  --overlay-content-background: var(--focus-background);
+  --overlay-content-background: var(--content-start-background);
```
dskloetd, line 115: As far as I understand, the meaning of `--content-start-background` is specifically the background color of the `start` section of the `SplitContent` component:
https://github.com/dfinity/gix-components/blob/367d349ed23f53f4fd8a5eff4a6f5e9e12229c1e/src/lib/components/SplitContent.svelte#L68

Is there as specific conceptual reason why we should reuse this color for the modal background? If not, would it make sense to define the color of `--overlay-content-background` directly (as `--cp-dark-450`) or to use a shared variable with a less specific meaning?

mstrasinskis: After giving it some more thought, I agree. Updated. Thanks!

dskloetd (approved):
Thanks!

PR #393, opened 2024-03-06
Add option to render popover without backdrop

dskloetd:
# Motivation

We want to render the popovers in NNS dapp without a backdrop.
Like in this design of the setting to hide tokens with zero balance:
https://www.figma.com/file/x8dk5axQUf8Bi43uUsqCHc/NNS-Production-%26-Hand-over?type=design&node-id=1529-6537&mode=design&t=7nLVmnqL7aLgPKCD-4
But also for the existing top-right menu.

We'll still need the backdrop to catch any clicks outside the popover, but we add an option to make it invisible.
The default behavior remains unchanged.

# Changes

1. Add prop `invisible` to backdrop, which prevents the `background` and `backdrop-filter` from being applied.
2. Add prop `invisibleBackdrop` (default `false`) to `Popover`, which is passed to the `Backdrop`.
3. Add a border to the popover if there is no backdrop to distinguish it from its surrounding.
4. Change the popover background color to what we have in Figma, so it has the correct contrast with the border color.

# Screenshots

With backdrop:
image

Without backdrop:
image


src/tests/lib/components/Popover.spec.ts
```diff
@@ -33,6 +33,22 @@ describe("Popover", () => {
       container.querySelector("div.backdrop");
 
     expect(backdrop).not.toBeNull();
+    expect(backdrop.getAttribute("class")).toContain(" visible");
```
mstrasinskis, line 36: What do you think, could using `split` improve the test to rely on the class position/quantity?
`expect(backdrop.getAttribute("class").split(' ')).toContain(" visible");`

src/tests/lib/components/Popover.spec.ts
```diff
@@ -33,6 +33,22 @@ describe("Popover", () => {
...
+        visible: true,
+        invisibleBackdrop: true,
+      },
+    });
+
+    const backdrop: HTMLDivElement | null =
+      container.querySelector("div.backdrop");
+
+    expect(backdrop).not.toBeNull();
+    expect(backdrop.getAttribute("class")).not.toContain(" visible");
```
daniel-wong-dfinity-org, line 51: Ditto.

src/tests/lib/components/Popover.spec.ts
daniel-wong-dfinity-org: I think you can say

```
e.classList.contains('foo')
```

daniel-wong-dfinity-org (commented):
I'm going to go over this with Maxims later, but I've read the code changes once, and saw a couple of possible micro improvements.

src/lib/components/Popover.svelte
```diff
@@ -99,6 +104,10 @@
     border-radius: var(--border-radius);
 
     box-shadow: var(--overlay-box-shadow);
+
+    &.with-border {
+      border: var(--tooltip-border-size) solid var(--tooltip-border-color);
```
peterpeterparker, line 109: the border color in the screenshot of the PR visually don't exactly match the outcome of Figma, the contrast isn't as visually distinctive. is that expected?

src/tests/lib/components/Popover.spec.ts
daniel-wong-dfinity-org: I think you can do

```
e.classList.contains('foo')
```

daniel-wong-dfinity-org (approved):

src/lib/components/Popover.svelte
dskloetd: The border color is exactly what it is in Figma, but the background color is not.
I was going to change the background color in another PR, but I've done it in this PR now.
Now it looks like this:
image

src/tests/lib/components/Popover.spec.ts
dskloetd: Yes, of course, good point. Done. Thanks.

src/tests/lib/components/Popover.spec.ts
dskloetd: Done

mstrasinskis (approved):
Thanks!

src/lib/components/Popover.svelte
peterpeterparker: Contrast is better I think, thanks!

PR #402, opened 2024-03-26
style: improve segment initial selection animation

mstrasinskis:
# Motivation

Improve segment initial selection animation.

# Changes

- do not render indicator before we know it's position.
- display segment buttons with the short opacity transition to avoid displaying before the indicator is rendered.

# Screenshots

| Before | After |
|--------|--------|
| Screen Recording 2024-03-26 at 15 47 09 | Screen Recording 2024-03-26 at 15 48 14  |


src/lib/components/SegmentButton.svelte
```diff
@@ -53,7 +58,14 @@
     padding: var(--padding-0_5x) var(--padding-1_5x);
 
     @include text.truncate;
-    transition: color var(--animation-time-normal);
+    transition:
+      color var(--animation-time-normal),
```
dskloetd, line 62: Does color need to be animated if opacity is already animated?
And why to they have a different animation time?

dskloetd (commented):
After reading the PR description I'm not sure what causes the problem and why you chose this fix.

First, if I look at the "before" animation, it looks like the selected button is already known from the start because it has white text instead of black text. So I'm not even sure any kind of animation is required. Can it just be rendered correctly from the start?

I'd also like to know why we need to know whether the component is mounted. Is there really something else that needs to be true to be able to correctly render the component and it just so happens to be the case that this is true once the component is mounted? If so, it would be good to check that other thing explicitly.

src/lib/components/SegmentButton.svelte
mstrasinskis: Color for selection, opacity for initial animation.

mstrasinskis:
> After reading the PR description I'm not sure what causes the problem and why you chose this fix.

The complexity comes from the fact that the buttons are rendered in the slot.

dskloetd (approved):

peterpeterparker:
Have you try to animate the selector ("blue background") only and not the appearance of the text? Is it a bit smoother?

mstrasinskis:
> Have you try to animate the selector ("blue background") only and not the appearance of the text? Is it a bit smoother?
> 
> Or maybe displaying the text with an opacity like 0.4 to 1 instead of 0 to 1.

Here is the version w/o opacity. And w/o applying selection before initialisation:

```
    transition: color var(--animation-time-normal);

    &.selected.initialised {
      color: var(--primary-contrast);
    }
```

Screen Recording 2024-03-27 at 09 55 11

That's good that all texts are initially displayed. Let's do like this, thanks!

peterpeterparker:
I like it better too, thanks!

dskloetd (approved):

PR #401, opened 2024-03-22
Remove box-shadow on Popover

dskloetd:
# Motivation

Artem asked me to remove the box-shadow on the Popover.
Until recently the Popover always had a visible dark backdrop where the box-shadow wasn't really visible anyway so this should only make a difference for the new cases where we hide the backdrop.

# Changes

Remove `box-shadow` style on `Popover`.

# Screenshots

Before:
image

After:
image

PR #267, opened 2023-08-02
fix: card spacing and alignment

peterpeterparker:
# Motivation

Card spacing and alignment is not 100% precise following introduction of the card links.

# Note

This PR also partially revert #266 to previous values.

# Changes

- play with margins

# Screenshots

Issues

issue-1
issue-2

After 

Capture d’écran 2023-08-02 à 11 40 45
Capture d’écran 2023-08-02 à 11 40 37
Capture d’écran 2023-08-02 à 11 40 12
Capture d’écran 2023-08-02 à 11 40 09
Capture d’écran 2023-08-02 à 11 40 04

PR #403, opened 2024-03-27
Project dropdown related adjustments

mstrasinskis:
# Motivation

`gix-components` changes for mobile project selector redesign. Part of https://github.com/dfinity/nns-dapp/pull/4668

# Changes

- `noMargin ` prop to remove margins from a `Card`. To avoid stretching of a `SplitContent` start block on mobile.
- `--nav-component-title-display` css-variable to control the nav title  to control it's visibility from outside.
- don't display `SplitContent` start background on mobile.

# Screenshots
| Before | After |
|--------|--------|
| image | image |
 |


src/lib/components/Card.svelte
```diff
@@ -15,6 +15,7 @@
     undefined;
   export let href: string | undefined = undefined;
   export let noPadding = false;
+  export let noMarginBottom = false;
```
dskloetd, line 18: I'm wondering wouldn't the standard way to do this be to have a `--card-margin-bottom` css variable that can be overwritten by the client?

I'm not against this but I think I've seen something like that so I'm wondering if there's any consistency that we can stick to.

src/lib/components/Card.svelte
```diff
@@ -112,6 +114,10 @@
       padding: 0;
     }
 
+    &.noMarginBottom {
+      margin-bottom: 0;
```
dskloetd, line 118: Do I see correctly on line 180 that if the card has an `href`, then it already has no margin?
Any idea why those 2 things are related?
Is it possible that we can just remove the margin in all cases rather than adding more and more special cases?

src/lib/components/Nav.svelte
```diff
@@ -9,18 +9,17 @@
   @use "../styles/mixins/media";
 
   div {
-    padding: var(--padding-2x) var(--padding-2x) 0;
-    margin: 0;
+    // title is hidden on small devices
```
dskloetd, line 12: Did you decide to make this change for all clients?

src/lib/components/SplitContent.svelte
```diff
@@ -62,15 +61,17 @@
...
   .start {
     min-height: var(--content-start-height);
-
-    background: var(--content-start-background);
     color: var(--content-start-color);
 
     @include media.min-width(large) {
       display: block;
       width: var(--content-start-width);
+      background: var(--content-start-background);
```
dskloetd, line 74: Do you need both line 65 and line 74?

src/lib/components/Card.svelte
mstrasinskis: Here we remove only a bottom margin. The top margin is still needed. Not sure it worth the risk, the card is so generic and widely used component.

src/lib/components/Card.svelte
mstrasinskis: Then it should be at least `--card-article-margin-bottom`. I think only having different components for different card types could simplify it.

src/lib/components/Nav.svelte
mstrasinskis: Yes, pretty bold move to avoid extra complexity. The solution with `{#if $slot.title...` wouldn't work, because it depends on the breakpoint and it's not allowed to wrap `<div slot="...` into conditional operators.

src/lib/components/SplitContent.svelte
mstrasinskis: Weird, 64-66 should be removed.

mstrasinskis: Good catch, was removed.

src/lib/components/Card.svelte
dskloetd: The possibility to add an `href` to the card and have it be an `<a>` instead of an `<article>` as a result was only added last August (in [this PR](https://github.com/dfinity/gix-components/pull/262)) and removing the margin for those cards was added on the same day (in [this PR](https://github.com/dfinity/gix-components/pull/267)).

I think it's pretty bad to have the styling change based on whether there is a link. And I think it's worse to have multiple ways to remove different parts of the margin. I'm afraid will end up with 4 different props for all different margins.

Instead of having a prop for just the bottom margin, can we have a prop that removes all the margins? If we need a top margin, we can add it back in nss-dapp.

Ideally we would then remove the special styling for `<a>` cards and apply the same prop to get the same effect where needed. But we can think about that separately.

e2e/popover.e2e.ts-snapshots/Should-render-close-button-1-Google-Chrome-linux.png
dskloetd, line 1: FYI, this change should have been in https://github.com/dfinity/gix-components/pull/401 but somehow it wasn't caught there. It's OK to add it here. Thanks.

src/lib/components/Nav.svelte
```diff
@@ -1,4 +1,4 @@
-<div><slot name="title" /></div>
+<div class="nav-title"><slot name="title" /></div>
```
dskloetd, line 1: How does this let you control visibility from outside?
Does it require `:global` or something like that? I think we should try to avoid that.
If I'm not mistaken, the proper way is to add a style which uses a CSS variable and then you can use the variable to change the style from outside.

mstrasinskis: You're right, it requires using `:global` https://github.com/dfinity/nns-dapp/pull/4668/files#diff-243e2f8111bf54358b075a4d7f13db58aa592be9394210efe583fa413326d4faR28. It can be also solved w/ the css-variable, but to have the possibility to rewrite the css variable, it should also be somewhere outside of the `Nav` component (for example in `variables.scss`). And having `--nav-component-title-display: block;` outside of the `Nav` component sounds a bit weird to me, but I'm happy to discuss this. Unconvincing excuse: the current solution serves as a temporary state until we have the SplitContent redesign.

src/lib/components/Card.svelte
mstrasinskis: This sounds as a lot of refactoring. Let's do it separately. Good investigation btw 👍

src/lib/components/Nav.svelte
mstrasinskis: The title is back.

src/lib/components/Card.svelte
dskloetd: It's not a lot of refactoring. All I'm asking is that you change `noMarginBottom` to `noMargin` and that it removes all the margins instead of just the bottom margin.

src/lib/components/Nav.svelte
dskloetd: I think it's good that something needs to be exposed explicitly, rather than through `:global`.
It can be very difficult to find out why something is missing if it's not explicit.
Can we please do the CSS variable?

mstrasinskis: Let's use css-var then. `--nav-component-title-display` was added.

src/lib/components/Card.svelte
mstrasinskis: Done, thanks.

src/lib/components/Card.svelte
mstrasinskis: Used `noMargin` instead 👍

dskloetd (approved):

PR #404, opened 2024-03-28
style: fix Segment mobile width

mstrasinskis:
# Motivation

The segment button uses more space then it shoudl on mobile.

# Changes

- restore using css-vars for segment width that were removed during restyling (https://github.com/dfinity/gix-components/pull/388/files#diff-2212548ca2a70ca70d0425f28f85ab2e91ef851bf1c0db9e0dc3fa740359bb55L99)
- add `box-sizing: border-box;` to respect segment padding (screenshots);

# Screenshots

| Before | After |
|--------|--------|
|Screenshot 2024-03-28 at 17 33 05 | Screenshot 2024-03-28 at 17 35 47 |


src/lib/components/Segment.svelte
```diff
@@ -105,8 +105,6 @@
     border-radius: var(--border-radius);
     overflow: hidden;
 
-    width: 100%;
```
dskloetd, line 108: Can you investigate why this was here an explain why it's no longer necessary?

src/lib/components/Segment.svelte
```diff
@@ -105,8 +105,6 @@
...
-
     @include media.min-width(medium) {
       width: var(--segment-min-width, var(--segment-width, fit-content));
```
dskloetd, line 109: On desktop we set a special width? Don't we need any kind of width on mobile?

src/lib/components/Segment.svelte
mstrasinskis: One part of the Segment restyling was to use whole width on mobile, instead of calculated width. And this rule was added by mistake because it's not necessary and lead to increased width because of the padding. 
`width: 100%;` with `box-sizing: border-box;` could also fix it, but better use less css then more when the result is the same.

src/lib/components/Segment.svelte
mstrasinskis: Let's set it explicitly w/ css-vars https://github.com/dfinity/nns-dapp/pull/4674

src/lib/components/Segment.svelte
```diff
@@ -105,12 +105,9 @@
...
     overflow: hidden;
 
-    width: 100%;
-
-    @include media.min-width(medium) {
-      width: var(--segment-min-width, var(--segment-width, fit-content));
-      --segment-min-width: calc(var(--segment-button-width) * var(--segments));
-    }
+    box-sizing: border-box;
+    width: var(--segment-min-width, var(--segment-width, fit-content));
```
lmuntaner, line 109: Why do you need a fallback for `--segment-min-width` if it's defined below?

src/lib/components/Segment.svelte
```diff
@@ -105,12 +105,9 @@
...
+    --segment-min-width: calc(var(--segment-button-width) * var(--segments));
```
lmuntaner, line 110: Does it work defining it after being used?

lmuntaner (approved):
Rubber stump but adding some questions.

src/lib/components/Segment.svelte
mstrasinskis: Seems so. GPT: In CSS, the order in which you define CSS variables (also known as custom properties) relative to their usage does not generally affect their functionality, thanks to CSS's cascading and scope rules. CSS variables are accessible within the scope they are declared, and they can be used before they are defined in the document flow. This is because CSS parsing does not strictly depend on the order of declaration as it does in some programming languages like JavaScript.

src/lib/components/Segment.svelte
mstrasinskis: I think the declaration place of `--segment-min-width` is not important if in the same scope. If the `--segment-button-width` is not declared (in nns-dapp for example), the auto-size mechanism is being applied.

PR #405, opened 2024-04-01
fix: wait for segment ready

mstrasinskis:
# Motivation

Fix segment indicator position when not the first element is selected.

# Changes

- wait for component ready before using DOM element sizes.

# Screenshots

| Before | After |
|--------|--------|
| Screen Recording 2024-04-01 at 22 40 57 | Screen Recording 2024-04-01 at 22 37 30 |


dskloetd (commented):
I don't understand the cause of the problem or the fix.
I hope you can explain it to me when you are back from vacation.

mstrasinskis:
> I hope you can explain it to me when you are back from vacation.

@dskloetd, the banner - "You are not signed in" is displayed only when the "Actionable Proposals" are selected (second position of the segment). The issue was that after switching between universes, the selection always pointed to the first segment button.

src/lib/components/Segment.svelte
```diff
@@ -44,6 +44,11 @@
       return;
     }
 
+    // wait for segment buttons take their place
+    if (isNullish(indicator)) {
+      await tick();
```
dskloetd, line 49: I've checked out your branch to try for myself. The problem isn't that the selection is pointing to the first button. It's actually pointing to the second button, but because `currentClientWidth` is equal to 4, it only moves the selection 4 pixels to the right to point to the second button.
I thought maybe the measurement is wrong because the element isn't in the DOM, but if I output all its ancestors, it is inside the main document.
So I'm not actually sure why it measures 4 pixels wide initially.

I'm going to approve to unblock you, but I don't think `tick` is a proper way to do this. Have you tried waiting for the component to be mounted before calling `initIndicator`?

Also, please fix the comment on line 47 to something like
```
// Initially selectedElement only has a width of 4 pixels, putting the indicator in the wrong position.
// Wait in order to get a proper measurement of selectedElement.
```

Also, have you tried if this is reproducible in a unit test?

dskloetd (approved):

src/lib/components/Segment.svelte
dskloetd: Sorry, my explanation was also wrong.
The `indicator.width` is actually completely ignored so it's irrelevant that it's 4 pixels.
The problem is that `currentClientLeft - parentClientLeft` is only 8 pixels.

PR #411, opened 2024-04-17
chore: prevent console log usage with eslint

peterpeterparker:
# Motivation

Prevent usage of `console.log` with eslint (given that I spotted a console.log in a test while developing an unrelated feature).


mstrasinskis (approved):
Thanks!

PR #412, opened 2024-04-17
chore: solve lint error type 'undefined' is not assignable to type 'Node'

peterpeterparker:
# Motivation

Lint currently detects an error on `main`:

```
/Users/daviddalbusco/projects/dfinity/gix-components/src/lib/components/Tooltip.svelte:69:31
Error: Argument of type 'HTMLDivElement | undefined' is not assignable to parameter of type 'Node'.
  Type 'undefined' is not assignable to type 'Node'. (ts)
    // Move tooltip to the body to avoid it being cut off by overflow: hidden.
    document.body.appendChild(tooltipComponent);
  });
```


mstrasinskis (approved):
LGTM, thank you!

PR #419, opened 2024-05-06
Close modal on esc

mstrasinskis:
# Motivation

To improve use experience the modal can be closed also on pressing `Esc` button.

# Changes

- Listen for keydown and close.

# Test

- Tested manually using upgraded `nns-dapp`.
- Added a test.

# Screenshots

No visual changes.

### DIFF
--- a/src/lib/components/Modal.svelte
+++ b/src/lib/components/Modal.svelte
@@ -30,7 +30,15 @@
 
   const modalTitleId = nextElementId("modal-title-");
   const modalContentId = nextElementId("modal-content-");
+
+  const handleKeyDown = (event: KeyboardEvent) => {
+    if (visible && !disablePointerEvents && event.key === "Escape") {
+      close();
+    }
+  };
 </script>
+
+<svelte:window on:keydown={handleKeyDown} />
 
 {#if visible}
   <div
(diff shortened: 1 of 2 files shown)


src/lib/components/Modal.svelte
```diff
@@ -30,8 +30,16 @@
 
   const modalTitleId = nextElementId("modal-title-");
   const modalContentId = nextElementId("modal-content-");
+
+  const handleKeyDown = (event: KeyboardEvent) => {
```
peterpeterparker, line 34: I once read it was useful to not name event as such because once bundle it might conflicts with event if polyfied. Long story short, `{key}`?

src/tests/lib/components/Modal.spec.ts
```diff
@@ -123,6 +123,19 @@ describe("Modal", () => {
       backdrop && fireEvent.click(backdrop);
     }));
 
+  it("should trigger close modal on Esc", () =>
```
peterpeterparker, line 126: Following one suggestion dK made me in another PR, worth adding a test that the modal does not close if one press another key like 'a' or 'b' ?

peterpeterparker (approved):
Yes, awesome! Thanks!  🥳

src/tests/lib/components/Modal.spec.ts
mstrasinskis: I did it manually, but agree, let's have it in the list.

PR #421, opened 2024-05-09
build: update Juno setup

peterpeterparker:
# Motivation

Bring the cool stuffs.

# Changes

- Upgrade analytics lib for breaking changes of last Juno release
- Analytics lib is a dependency, not dev dependency
- Transform json config into a TypeScript config file and add types to add intelisense in the editor
- Use Juno vite plugin to read and inject IDs instead of duplicating the values in env file


lmuntaner (approved):
Rubber stamp

PR #422, opened 2024-05-09
chore: modal lint error

peterpeterparker:
# Motivation

Fix lint error:

> /Users/daviddalbusco/projects/dfinity/gix-components/src/lib/components/Modal.svelte:35:28
> Error: Binding element 'key' implicitly has an 'any' type. (ts)
>
>  const handleKeyDown = ({ key }) => {
>    // Check for $busy to mock the same behavior as the close button being covered by the busy overlay


dskloetd (approved):
Thanks!

PR #424, opened 2024-05-19
feat: VoteIcon size param

mstrasinskis:
# Motivation

We need to display a vote icon in a select actionable button at a non-standard size.

# Changes

- Add a `size` prop (tested manually).

# Screenshots

image


peterpeterparker (approved):
LGTM, thx

PR #427, opened 2024-05-28
Use get instead of $

mstrasinskis:
# Motivation

There is a bug that causes nns-dapp to crash. It’s possible to reproduce it locally against the dev version. It seems to be a weird Svelte behavior during unsubscribing from the store inside the window listener, but no real proof of this has been found. However, removing the store subscription leads to normal nns-dapp behavior without crashes.

How to reproduce the bug:

1. Sign into https://nns.ic0.app/
2. Click on Internet Computer
3. Click on “Receive”
4. Click on “Finish”
5. The site crashes.

# Changes

- Replacing `$` with `get(store)`.

# Test

- Manually tested.


dskloetd (approved):
Thanks!

PR #430, opened 2024-06-07
Add idPrefix prop to Tooltip

dskloetd:
# Motivation

Every tooltip must have a unique ID so the target can be correctly associated with the tooltip through `describedby`.
Giving a unique ID to a tooltip in a component that appears multiple times on the page, can be tedious.

# Changes

1. Add `idPrefix` prop to `Tooltip` and make `id` optional. If `id` is not specified, a unique ID is automatically generated by combining the prefix with a sequence number. For backwards compatibility, if `id` is specified, it's used as is.

# Tested

1. Unit tests were added.
2. Tested in nns-dapp repo.

### DIFF
--- a/src/lib/components/Tooltip.svelte
+++ b/src/lib/components/Tooltip.svelte
@@ -1,9 +1,14 @@
+<script lang="ts" context="module">
+  let nextTooltipIdSuffix = 0;
+</script>
+
 <script lang="ts">
   import { onMount, onDestroy } from "svelte";
   import { nonNullish, notEmptyString } from "@dfinity/utils";
   import { translateTooltip } from "$lib/utils/tooltip.utils";
 
-  export let id: string;
+  export let id: string | undefined = undefined;
+  export let idPrefix = "tooltip";
   export let testId = "tooltip-component";
   export let text: string | undefined = undefined;
   export let noWrap = false;
@@ -16,6 +21,8 @@
   let tooltipTransformX = 0;
   let tooltipTransformY = 0;
   let tooltipStyle: string | undefined = undefined;
+
+  let idToUse = nonNullish(id) ? id : `${idPrefix}-${nextTooltipIdSuffix++}`;
 
   $: tooltipStyle = `--tooltip-transform-x: ${tooltipTransformX}px; --tooltip-transform-y: ${tooltipTransformY}px;`;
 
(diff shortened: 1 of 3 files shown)


src/tests/lib/components/Tooltip.spec.ts
```diff
@@ -16,15 +19,59 @@ describe("Tooltip", () => {
...
+    expect(describedBy2).toMatch(new RegExp(`^${idPrefix}-\\d+$`));
+    expect(
+      container.querySelector(`[id="${describedBy1}"]`),
+    ).toBeInTheDocument();
+    expect(
+      container.querySelector(`[id="${describedBy2}"]`),
+    ).toBeInTheDocument();
+  });
+
+  it("should render different ID per tooltip when not specifying any ID", () => {
```
lmuntaner, line 47: What is the difference between this test and the previous one?

I only see that in this one you don't check against the regex.

src/lib/components/Tooltip.svelte
```diff
@@ -1,9 +1,14 @@
+<script lang="ts" context="module">
+  let nextTooltipIdSuffix = 0;
```
lmuntaner, line 2: One question, if a tooltip is mounted and then unmounted, does it use a new `nextTooltipIdSuffix`?

lmuntaner (approved):
Two small questoins, but already approving.

src/tests/lib/components/Tooltip.spec.ts
dskloetd: Oops, this was supposed to not pass the `idPrefix` either. Fixed. Thanks.

src/lib/components/Tooltip.svelte
dskloetd: If it's a new instance of the component, it will get a new id.
I don't think the same instance can be mounted multiple times (can it?) but if it could, it would have the same id.

lmuntaner: So this number will go up for every new instance. If a page has 10 tooltips it go up to 10, the user navigates away, tooltips are dismounted and when the user comes back, 10 more are rendered with ids 11 to 20, right?

dskloetd: Yes, well, only for those tooltips that don't specify in `id`.
Does having the number constantly increase seem concerning to you?

lmuntaner: Not at all. I wanted to understand the functionality only.

dskloetd: OK, thanks!

PR #438, opened 2024-06-17
Support rich content in tooltip

dskloetd:
# Motivation

Currently the contents of the tooltip is specified as a text string.
We want to be able to show richer tooltips with styling.

# Changes

Allow passing the tooltip content as a slot instead of just a `text` prop.

Note that with the `text` prop it's currently possible to disable the tooltip entirely by making the `text` prop empty or `undefined`.
When using the slot, there is currently no way to make the tooltip conditional.
This would require an additional boolean prop.
This might be useful but we'll add it when we need it.

# Screenshots

image

### DIFF
--- a/src/lib/components/Tooltip.svelte
+++ b/src/lib/components/Tooltip.svelte
@@ -3,8 +3,8 @@
 </script>
 
 <script lang="ts">
+  import { isNullish, nonNullish, notEmptyString } from "@dfinity/utils";
   import { onMount, onDestroy } from "svelte";
-  import { nonNullish, notEmptyString } from "@dfinity/utils";
   import { translateTooltip } from "$lib/utils/tooltip.utils";
 
   export let id: string | undefined = undefined;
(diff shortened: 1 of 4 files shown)


src/lib/components/Tooltip.svelte
```diff
@@ -102,12 +105,12 @@
...
     class:noWrap
     class:top
-    class:not-rendered={!notEmptyString(text)}
+    class:not-rendered={isAbsent}
     class:visible={targetIsHovered}
     bind:this={tooltipComponent}
     style={tooltipStyle}
   >
-    {text}
+    {#if nonNullish(text)}{text}{/if}<slot name="tooltip-content" />
```
mstrasinskis, line 113: Will it be more predictable to use if else here?

mstrasinskis (approved):
Nice. Thanks!

src/lib/components/Tooltip.svelte
dskloetd: I wouldn't expect the user to use both the `text` prop and the slot, but if they do I think it should either show both or cause an error. I don't think we should hide one or the other silently.

PR #444, opened 2024-06-19
Add resetScrollPosition function

mstrasinskis:
# Motivation

When switching between universes, the user would find themselves somewhere in the middle of the list. This issue went unnoticed before due to paginated loading, which resulted in the list being empty immediately after switching. However, with actionables now being preloaded, this has become a significant UX issue. 

The proposals are rendered inside the `SplitContent` component. To reset the scroll position without having to search through the DOM, we have added a `resetScrollPosition` function.

# Changes

- new `resetScrollPosition` function.

# Screenshots

No visual changes. Tested in nns-dapp.

### DIFF
--- a/src/lib/components/SplitContent.svelte
+++ b/src/lib/components/SplitContent.svelte
@@ -9,6 +9,13 @@
   import ContentBackdrop from "$lib/components/ContentBackdrop.svelte";
 
   export let back = false;
+  export const resetScrollPosition = () => {
+    if (scrollableElement) {
+      scrollableElement.scrollTop = 0;
+    }
+  };
+
+  let scrollableElement: HTMLElement | undefined;
 
   // Same as in <Content />
   onDestroy(() => ($layoutBottomOffset = 0));
@@ -33,7 +40,7 @@
       <slot name="toolbar-end" slot="toolbar-end" />
     </Header>
 
-    <div class="scrollable-content-end">
+    <div class="scrollable-content-end" bind:this={scrollableElement}>
       <ContentBackdrop />
 
       <slot name="end" />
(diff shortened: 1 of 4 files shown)


src/lib/components/SplitContent.svelte
```diff
@@ -9,6 +9,13 @@
   import ContentBackdrop from "$lib/components/ContentBackdrop.svelte";
 
   export let back = false;
+  export const resetScrollPosition = () => {
+    if (scrollableElement) {
+      scrollableElement.scrollTop = 0;
+    }
+  };
+
+  let scrollableElement: HTMLElement | undefined;
```
dskloetd, line 18: Is the `| undefined` necessary?

src/lib/components/SplitContent.svelte
dskloetd, line 1: Can you add a test?

src/lib/components/SplitContent.svelte
mstrasinskis: Just for formality.  For example if we call `resetScrollPosition` on the next line, `scrollableElement` will be undefined.

dskloetd: Have you tried that? I'm really not sure but I thought maybe because the `div` below is rendered unconditionally, this might always be defined.

src/lib/components/SplitContent.svelte
mstrasinskis: Added https://github.com/dfinity/gix-components/pull/444/files#diff-4146bf35709d43bf5d5ffa125b632e3ad1b8a1dc0d08bd2deb29525378a2e9c1R27

src/lib/components/SplitContent.svelte
mstrasinskis: Yep, I tried.

dskloetd: Thanks!

dskloetd (approved):

PR #449, opened 2024-06-20
fix: Tooltip showcase page unique ids

peterpeterparker:
# Motivation

Ids must be unique within an HTML page. On the Tooltip showcase page, we are using multiple time the selement identitifier.

# Changes

- Increment example identitfiers.

### DIFF
--- a/src/routes/(split)/components/tooltip/+page.md
+++ b/src/routes/(split)/components/tooltip/+page.md
@@ -65,7 +65,7 @@ The tooltips will appear when the buttons are hovered or tapped.
     <div class="row">
       <Tooltip
         id="example-button"
-        text="text"
+        text={tooltipText}
       >
         <button class="secondary" disabled>Disabled</button>
       </Tooltip>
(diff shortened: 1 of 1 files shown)


src/routes/(split)/components/tooltip/+page.md
```diff
@@ -70,35 +70,35 @@ The tooltips will appear when the buttons are hovered or tapped.
...
       <Tooltip
-        id="example-button"
+        id="example-button-1"
         text={tooltipText}
       >
         <button class="secondary" disabled>Disabled</button>
       </Tooltip>
       <Tooltip
-        id="example-button"
+        id="example-button-2"
```
dskloetd, line 79: FYI, another way to accomplish the same is to use
```
idPrefix="example-button"
```

dskloetd (approved):
Thanks!

src/routes/(split)/components/tooltip/+page.md
peterpeterparker: Good to know!

PR #450, opened 2024-06-21
chore: do not accept lint warnings in CI

peterpeterparker:
# Motivation

We do not want the CI to pass when `npm run lint` detects warnings.


dskloetd (approved):
Thanks!

PR #453, opened 2024-06-24
Move tooltip back to body in afterUpdate

dskloetd:
# Motivation

I noticed in the neurons table that when the order of the rows is changed, the tooltips are moved inside the `overflow: hidden` container and end up being cut off.
We move tooltips directly to the document body in `onMount` to make sure they are not cut off.
The solution I found is to move the tooltips again in `afterUpdate`.

# Changes

Move tooltips to the document body in `afterUpdate` the same as in `onMount`.

# Tested

1. Unit test added.
2. Checked that the unit test failed with `afterUpdate` commented out.
3. Updated e2e test because the tooltips are no longer guaranteed to be in the same order in the DOM as the tooltip targets.
4. Tested manually in an nns-dapp branch.

### DIFF
--- a/src/lib/components/Tooltip.svelte
+++ b/src/lib/components/Tooltip.svelte
@@ -4,7 +4,7 @@
 
 <script lang="ts">
   import { isNullish, nonNullish, notEmptyString } from "@dfinity/utils";
-  import { onMount, onDestroy } from "svelte";
+  import { afterUpdate, onMount, onDestroy } from "svelte";
   import { translateTooltip } from "$lib/utils/tooltip.utils";
 
   export let id: string | undefined = undefined;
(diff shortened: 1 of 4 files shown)


src/lib/components/Tooltip.svelte
```diff
@@ -74,10 +74,14 @@
...
-  onMount(async () => {
+  const moveTooltipToBody = () => {
     // Move tooltip to the body to avoid it being cut off by overflow: hidden.
     nonNullish(tooltipComponent) && document.body.appendChild(tooltipComponent);
-  });
+  };
+
+  onMount(moveTooltipToBody);
+
+  afterUpdate(moveTooltipToBody);
```
mstrasinskis, line 84: Just thinking out loud, would it help to call `moveTooltipToBody` only on hover?

mstrasinskis (approved):
Thank you!

src/lib/components/Tooltip.svelte
dskloetd: If they're not moved, they actually take up space and cause scroll bars where there shouldn't be.
I suppose there may be a way to solve that too. Feel free to give it a try. Let me know if you need help reproducing.

PR #460, opened 2024-06-30
New icon plus

mstrasinskis:
# Motivation

A new icon is needed for the import token button.

# Changes

- New plus icon.

# Screenshots

image


dskloetd:
We have this UI for adding an account:
image

Don't we want to be consistent with that? Or are we going to use the icon there as well?

mstrasinskis:
> We have this UI for adding an account: image
> 
> Don't we want to be consistent with that? Or are we going to use the icon there as well?

Good point. That icon will be updated as well.

dskloetd (approved):

PR #462, opened 2024-07-01
chore: update local snapshots

peterpeterparker:
# Motivation

This PR bulk update the snapshots that can be used locally on Mac when running the e2e tests locally.

PR #463, opened 2024-07-01
chore: remove local screenshots

peterpeterparker:
# Motivation

As discussed in #462, we do not need the local screenshots in Git.

# Changes

- Remove existing screenshots.
- Add a `.gitignore` to ignore screenshots generated on darwin (we all use Mac).
- Update hacking doc to remove local screenshots steps.


dskloetd (approved):
Thanks

PR #465, opened 2024-07-17
Remove buttons default padding

mstrasinskis:
# Motivation

Currently, a `button` tag by default has a small (0.5rem) horizontal padding. This breaks the alignment in the account menu because some entries are rendered with an `a` tag, which has no extra padding, while others use a `button` tag.

There are not so many affected buttons. For many buttons the padding is overwritten: g.e.`.primary, .secondary, .success, .danger, .warning` (based on the `button.base` mixin), `icon-only`, 'icon'.

The buttons that are affected by the changes in this PR: `.full-width`, `.ghost`, `.text`, `.with-icon`, `.card`.

# Changes

- Set the padding 0 by default for buttons. This will not affect most of the buttons, only that who uses no @button.base mixin (copy button, 
- Remove `padding: 0` from the places where the default padding is not needed.

# Screenshots

### DIFF
--- a/src/lib/components/Collapsible.svelte
+++ b/src/lib/components/Collapsible.svelte
@@ -147,7 +147,6 @@
     width: var(--padding-4x);
 
     margin: 0;
-    padding: 0;
 
     display: flex;
     align-items: center;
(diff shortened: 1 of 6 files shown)


mstrasinskis:
Closing this PR as it’s no longer relevant.

Issue #469, opened 2024-08-02
sass nested rules breaking change


peterpeterparker:
# Issue

Sass `1.77.8` introduces a breaking changes that requires declarations to appear before nested rules. Two developers using gix-cmp have reported the issue as they upgraded this library and sass at the same time.

# Solution

We should bump sass in gix-cmp and fix the required changes - sort the declarations and nested rules.

# Screenshot

![image](https://github.com/user-attachments/assets/b783f345-7cc3-4294-b863-65320631ae2a)

PR #471, opened 2024-08-05
feat: upgrade sass

peterpeterparker:
# Motivation

Resolve issue #469 reported by developers who are using gix-cmp.

# Background

Sass 1.77.8 introduces a deprecation warning:

```
Deprecation Warning: Sass's behavior for declarations that appear after nested
rules will be changing to match the behavior specified by CSS in an upcoming
version. To keep the existing behavior, move the declaration above the nested
rule. To opt into the new behavior, wrap the declaration in `& {}`.

More info: https://sass-lang.com/d/mixed-decls

    ╷
10  │ ┌   &:active {
11  │ │     @if $color != null {
12  │ │       color: $color;
13  │ │     }
14  │ │     background-color: var(--tertiary);
15  │ │     background-size: 100%;
16  │ │     transition: background 0s;
17  │ │   }
    │ └─── nested rule
... │
20  │     transition: background 0.8s;
    │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ declaration
    ╵
    src/lib/styles/mixins/_effect.scss 20:3                  ripple-effect()
    src/lib/styles/mixins/_header.scss 21:5                  button()
    src/lib/components/Back.svelte.vite-preprocess.scss 6:5  root stylesheet

Warning: 4 repetitive deprecation warnings omitted.
```

# Changes

- Update Sass.
- Resolve warnings my moving declarations above rules and using `& {}` in two mixins to preserve current behavior.


src/lib/styles/mixins/_effect.scss
```diff
@@ -1,4 +1,9 @@
 @mixin ripple-effect($background, $color: null) {
+  & {
```
dskloetd, line 2: What does `& {` do any why is it necessary?

src/lib/styles/mixins/_header.scss
```diff
@@ -3,15 +3,13 @@
 @use "./media";
 
 @mixin button($background) {
-  width: var(--padding-4x);
-  min-width: var(--padding-4x);
-  height: var(--padding-4x);
+  & {
```
dskloetd, line 6: Why did you move the declarations inside `& {`?

src/lib/styles/mixins/_effect.scss
peterpeterparker: `& {` is equals to opting out to the waring - i.e. preserving current behavior.

```
To keep the existing behavior, move the declaration above the nested
rule. To opt into the new behavior, wrap the declaration in `& {}`.
```

I can add a TODO but resolving the conflict that way in a mixin that is used in several places was more straightforward.

We can also take more time and check any usages if you prefer?

src/lib/styles/mixins/_header.scss
peterpeterparker: Same as https://github.com/dfinity/gix-components/pull/471#discussion_r1703744381

src/lib/styles/mixins/_effect.scss
peterpeterparker: Updated the description.

dskloetd: Don't we already opt in to the new behavior by increasing the version of Sass that we're using?

dskloetd: And why do we actually want to opt in to the new behavior in these 2 cases?

> We can also take more time and check any usages if you prefer?

If you're not confident that the changes are correct, then we should check.

peterpeterparker: > And why do we actually want to opt in to the new behavior in these 2 cases?

That way you don't have to go through all components that use the mixin and resolve manually the warning by modifying the order of the declarations and the rules.

> If you're not confident that the changes are correct

I'm confident it's correct. `& {}` has the same result as the style on `main`. 

```
.something {
   color: black;
}

===

.something {
   & {
      color: black;
   }   
}
```

dskloetd (approved):

PR #480, opened 2024-09-12
New IconUnknownToken

mstrasinskis:
# Motivation

If the token logo fails to load, it won’t be available. In such cases, a question mark icon will be displayed in place of the logo. This PR introduces the new question mark icon for this scenario.

# Changes

- Add new icon `IconUnknownToken`.

# Screenshots

image


dskloetd:
We already have code to use a question mark logo when the token logo is unknown:
https://github.com/dfinity/nns-dapp/blob/8615efce8254f212b93a031a2bed708322656e1d/frontend/src/lib/derived/icrc-universes.derived.ts#L35

Isn't it easier to change that image to match what we want then to have to use it via a component?

PR #482, opened 2024-09-17
update theme toggle and tests

coskucinkilic:
# Motivation

Update theme toggle according to new figma designs

# Changes

ThemeToggle is now a button instead of an input
Tests has been updated accordingly

# Screenshots
Screenshot 2024-09-17 at 14 16 49
Screenshot 2024-09-17 at 14 15 09

### DIFF
--- a/src/lib/components/ThemeToggle.svelte
+++ b/src/lib/components/ThemeToggle.svelte
@@ -1,54 +1,40 @@
 <script lang="ts">
-  import Toggle from "./Toggle.svelte";
   import { Theme } from "$lib/types/theme";
   import { themeStore } from "$lib/stores/theme.store";
   import { i18n } from "$lib/stores/i18n";
   import IconLightMode from "$lib/icons/IconLightMode.svelte";
   import IconDarkMode from "$lib/icons/IconDarkMode.svelte";
 
-  const switchTheme = ({ detail }: CustomEvent<boolean>) =>
-    themeStore.select(detail ? Theme.DARK : Theme.LIGHT);
+  const switchTheme = () => {
+    themeStore.select($themeStore === Theme.LIGHT ? Theme.DARK : Theme.LIGHT);
+  };
 
-  let checked: boolean;
-  $: checked = $themeStore !== Theme.LIGHT;
+  $: isDarkMode = $themeStore === Theme.DARK;
 </script>
 
-<div class="theme-toggle" data-tid="theme-toggle">
-  <div class="toggle">
+<button
+  data-tid="theme-toggle"
+  class="theme-toggle icon-only"
+  on:click={switchTheme}
+  aria-label={$i18n.theme.switch_theme}
+>
+  {#if isDarkMode}
     <IconLightMode />
-    <Toggle
-      bind:checked
-      on:nnsToggle={switchTheme}
-      ariaLabel={$i18n.theme.switch_theme}
-    />
+  {:else}
     <IconDarkMode />
-  </div>
-</div>
+  {/if}
+</button>
 
 <style lang="scss">
   .theme-toggle {
-    display: flex;
-    align-items: center;
+    color: var(--text-color);
+    padding: var(--padding);
+    background: var(--sidebar-button-background);
 
-    font-size: var(--font-size-h4);
-    line-height: var(--line-height-h4);
-
-    padding: 0 var(--padding-0_5x);
-    gap: var(--padding);
-  }
-
-  .toggle {
-    display: flex;
-    align-items: center;
-    grid-template-columns: repeat(3, auto);
-    grid-column-gap: 2px;
-
-    :global(svg) {
-      width: var(--padding-2x);
-      height: var(--padding-2x);
-
-      &:first-of-type {
-        margin-right: calc(var(--padding-0_5x) / 2);
+    &:hover {
+      background: var(--sidebar-button-background-hover);
+      :global(svg) {
+        fill: var(--text-color);
       }
     }
   }
--- a/src/tests/lib/components/ThemeToggle.spec.ts
+++ b/src/tests/lib/components/ThemeToggle.spec.ts
@@ -1,4 +1,6 @@
 import ThemeToggle from "$lib/components/ThemeToggle.svelte";
+import IconDarkMode from "$lib/icons/IconDarkMode.svelte";
+import IconLightMode from "$lib/icons/IconLightMode.svelte";
 import { themeStore } from "$lib/stores/theme.store";
 import { Theme } from "$lib/types/theme";
 import { fireEvent, render } from "@testing-library/svelte";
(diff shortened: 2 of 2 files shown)


src/lib/components/ThemeToggle.svelte
```diff
@@ -1,54 +1,40 @@
...
 
-  const switchTheme = ({ detail }: CustomEvent<boolean>) =>
-    themeStore.select(detail ? Theme.DARK : Theme.LIGHT);
+  const switchTheme = () => {
+    themeStore.select($themeStore === Theme.LIGHT ? Theme.DARK : Theme.LIGHT);
+  };
 
-  let checked: boolean;
-  $: checked = $themeStore !== Theme.LIGHT;
+  $: isDarkMode = $themeStore === Theme.DARK;
```
peterpeterparker, line 12: For consistency reason can you declare the variable.

```
let darkMode: boolean;
$: darkMode = $themeStore
```

NTH: I might also avoid naming it 'isSomething'. We generally prefer being expressive in our variable names, but I could be wrong.

peterpeterparker (commented):
Thanks for the PR. How does it look in the gix-cmp showcase? If it needs adaptation, are you going to provide those in another PR? (It’s probably best to do it in two separate PRs indeed.)

dskloetd:
Does this button work as a drop-in replacement of the toggle even if we don't move the component to a different place in the page?
If not, I think it would be better to introduce a new component instead of changing the existing one.

peterpeterparker:
> drop-in replacement 

Note that I left comments assuming it was a drop-in replacement (that the community dev using the lib should also adopt). 
Agree on introducing a new component if it isn't.

src/lib/components/ThemeToggle.svelte
coskucinkilic: Not sure about the naming conventions here but to me darkMode and isDarkMode means two different things.

src/lib/components/ThemeToggle.svelte
```diff
@@ -1,54 +1,40 @@
...
 
-<div class="theme-toggle" data-tid="theme-toggle">
-  <div class="toggle">
+<button
+  data-tid="theme-toggle"
+  class="theme-toggle icon-only"
+  on:click={switchTheme}
+  aria-label={$i18n.theme.switch_theme}
+>
+  {#if isDarkMode}
```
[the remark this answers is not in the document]
coskucinkilic: It displays one or the other based on isDarkMode. 

The button icon-only already has a ripple effect.

src/lib/components/ThemeToggle.svelte
```diff
@@ -1,54 +1,40 @@
...
-
-      &:first-of-type {
-        margin-right: calc(var(--padding-0_5x) / 2);
+    color: var(--text-color);
+    padding: var(--padding);
+    background: var(--sidebar-button-background);
+
+    &:hover {
+      background: var(--sidebar-button-background-hover);
+      :global(svg) {
```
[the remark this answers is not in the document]
coskucinkilic: The other option is to track the hover with js save it in a state use different icons for filled and outlined.

I thought this would be a better approach but happy to change it. Also let me know if there is a better way to conditionally render icons based on hover.

Ex: I can put two icons in each if block and based on hover i can change the displays if that is something you would prefer, but i think it is harder to read.

src/lib/components/ThemeToggle.svelte
coskucinkilic: it looks like it is used in just one other location, so not really common naming convention. Do you want me to change it?

https://github.com/dfinity/gix-components/blob/NNS1-2965-update-theme-toggle/src/lib/components/Tooltip.svelte#L28

coskucinkilic:
Waiting for an answer from Artem

src/lib/components/ThemeToggle.svelte
peterpeterparker: Naming was a NTH, works for me.

Declaring the variable would be consistent.

src/lib/components/ThemeToggle.svelte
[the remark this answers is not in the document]
peterpeterparker: I mean the appearance of the icon, like a fade or blur effect in and/or out.

src/lib/components/ThemeToggle.svelte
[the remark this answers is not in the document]
peterpeterparker: The svg can use a CSS variable which you assign on hover.

coskucinkilic:
Artem told me this is a replacement, please provide feedback accordingly

src/lib/components/ThemeToggle.svelte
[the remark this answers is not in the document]
coskucinkilic: I can for sure do that or modify svgs to take in a fill property, but i thought we would like to be more consistent with how we keep svgs. If some svgs support fill and others don't then it could create a problem.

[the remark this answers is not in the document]
peterpeterparker: Using a CSS variables seems less prone to issue to me than overriding global but, not a strong opinion neither. If it's fine for the colleagues, fine for me.

src/lib/components/ThemeToggle.svelte
dskloetd: I agree with Cosku, that `isDarkMode` is clearer. With `darkMode` I wouldn't be sure about the type but with `isDarkMode` it's obviously a `boolean`.

dskloetd:
> Artem told me this is a replacement, please provide feedback accordingly

I'm confused by this. Can you show me how this looks in the account menu instead of the current theme switcher in the account menu?

src/lib/components/ThemeToggle.svelte
coskucinkilic: Variable added.

src/lib/components/ThemeToggle.svelte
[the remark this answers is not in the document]
coskucinkilic: fade in added to icons.

src/lib/components/ThemeToggle.svelte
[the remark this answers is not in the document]
coskucinkilic: Added CSS variable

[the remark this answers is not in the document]
dskloetd: I would have expected multiple icons (with different display on hover) but a CSS variable in the icon sounds good to me.
We can add CSS variables to more icons as we need.
It's also fine with me if you want to update all the icons to support a fill variable (in another PR) but not necessary.

coskucinkilic:
@dskloetd here you go:
Screenshot 2024-09-17 at 15 42 31

dskloetd (approved):
Thanks!

src/lib/components/ThemeToggle.svelte
peterpeterparker: We tend to use `isSomething` for function and `something` for variable, there is not linter and it's just a feeling so for sure it's fine.

peterpeterparker:
Look weird in the context menu but, that's a question above my pay grade.

Can you also share the screenshots regarding my earlier [question](https://github.com/dfinity/gix-components/pull/482#pullrequestreview-2309596571) as well?

dskloetd:
@peterpeterparker is talking about this: https://github.com/dfinity/gix-components?tab=readme-ov-file#documentation--showcase

coskucinkilic:
@peterpeterparker 
Screenshot 2024-09-17 at 16 03 12
Screenshot 2024-09-17 at 16 03 19
@dskloetd there are couple of updates after your approval, could you please take a quick look?

peterpeterparker:
Thanks. In the context menu it's fine as well?

Capture d’écran 2024-09-17 à 16 08 08

coskucinkilic:
After talking with Artem again, he said we will most probably don't use it but as people are aready using it outside of NNS let's keep both toggle and the button.

I will be moving some files around, will leave the context menu as is with ThemeToggle

src/lib/styles/themes/dark.scss
```diff
@@ -146,6 +146,7 @@
   // Sidebar
   --sidebar-button-background: var(--cp-dark-1000-20);
   --sidebar-button-background-hover: var(--cp-dark-900);
+  --sidebar-icon: var(--text-color);
```
peterpeterparker, line 149: do we need this new global variable?

peterpeterparker (commented):
Can you edit the description of the PR, I think it does not match anymore given that you are providing a new component.

src/lib/styles/themes/dark.scss
coskucinkilic: We don't need it, I am trying to keep everything meaningfully in sync with Figma. I can use --text-color directly as well.

peterpeterparker: Ah ok I understand. I tend to not add global variable that may not be needed but if it fits Figma sure that's fine for me.

peterpeterparker (approved):
LGTM, thx

src/routes/(split)/components/theme-toggle-button/+page.md
```diff
@@ -0,0 +1,15 @@
...
+
+In addition to [theming](/styling/theming) we also provide an opinionated toggle to integrate easily a theme switcher in your dapp.
+
+```javascript
+<ThemeToggleButton />
+```
+
+## Showcase
+
+<ThemeToggleButton />
```
dskloetd, line 15: Not important at all, but I was wondering if it makes sense to put the switch and the button on the same show-case page so you can see them side by side and know the other one also exists if you look at one.

PR #488, opened 2024-09-27
delete hiding elements of menu on small screens

coskucinkilic:
# Motivation

Make logo and toggle of collapse menu visibile on small screens

# Changes

display nones have been removed.

# Screenshots

Screenshot 2024-09-27 at 14 51 17


dskloetd:
What happens if there's not enough space? Is it safe to do this before we add scrolling?

coskucinkilic:
I have jsut checked up to 450 px height everything is working fine, according to web most devices support this.
These are old smart phones i mean really old they still support it in landscape mode.

iPhone SE (1st generation): 640x1136 pixels, 4-inch screen
iPhone 5/5S/5C: 640x1136 pixels, 4-inch screen
Some older Android devices like Samsung Galaxy S4 Mini: 540x960 pixels, 4.3-inch screen

Here are some other devices which will display logo on top of menu items:

Nokia 3310 3G (2017 version): 240x320 pixels, 2.4-inch screen
Nokia 8110 4G: 240x320 pixels, 2.4-inch screen
Jio Phone (popular in India): 240x320 pixels, 2.4-inch screen

I am not sure if there is any need to support them, but if you want to be totally safe, I can add the correct media query to target these devices and not display it, however the rest of the app won't function on these devices as well. Personally I don't see any reason to hide anything, but please let me know your final feedback and i can update the logo accordingly. 

Menu items will turn into scroll when there is not enough high, the only question here is to hide or not hide logo under certain height? If we don't hide it, it can get rendered on top of menu items even thought they are scrollable on very small set of devices.

dskloetd (approved):
So the only risk is that the logo overlaps with the menu items? Then let's proceed, thanks!

dskloetd (changes requested):
Wait, this is gix-components, not nns-dapp.
How many phones this looks bad on depends on how many menu items the particular app has, right?
I also saw a PR specific to nns-dapp. Why do we need both PRs?

coskucinkilic:
We need both PRs because, menu collapse toggle and logo is control in gix-components. Menu footer and menu items are provided by nns-dapp. You are right if we have more menu items in the future, we will see the logo collapse on larger devices as well. There is still enough space for at least 2 more menu items.

dskloetd:
I thought we agreed that the right solution is to make the container scrollable.
Removing the `min-height` without making the container scrollable is just a quick fix to make the new buttons appear.
But we didn't get any bug report for the menu collapse or the logo so I don't think it's necessary to do a quick half-fix for them before we do the proper solution
Did I misunderstand or was there additional communication that I wasn't part of?

coskucinkilic:
That is what i understand from [this conversation thread](https://dfinity.slack.com/archives/C039M7YS6F6/p1727440012631669?thread_ts=1727437742.878599&cid=C039M7YS6F6). This is not a proper scroll solution, however it will make the toggle and logo appear in small screens.

PR #491, opened 2024-10-03
delete stopPropagation on onClick and onKeypress in Popover

coskucinkilic:
# Motivation

Make LinkToCanister component function as expected in nnsDapp

# Changes

stopPropagation on onClick and onKeypress in Popover are deleted

# Screenshots

### DIFF
--- a/src/lib/components/Popover.svelte
+++ b/src/lib/components/Popover.svelte
@@ -35,8 +35,8 @@
     style="--popover-top: {`${bottom}px`}; --popover-left: {`${left}px`}; --popover-right: {`${
       window.innerWidth - right
     }px`}"
-    on:click|stopPropagation
-    on:keypress|stopPropagation
+    on:click
+    on:keypress
   >
     <Backdrop
       on:nnsClose={() => (visible = false)}


peterpeterparker:
I'm not sure about this change. Doesn't it imply that, for example, if I click and select text displayed in a popover, the popover will close itself which we most likely not want?

Also, can you explain why this change was made? What issue does it resolve?

dskloetd:
@peterpeterparker it appears that this component originates from papyrs and the `stopPropagation` was added by you in 
https://github.com/papyrs/papyrs/commit/c80cf288d479d82461e583b8727195df5f200e89

It doesn't seem to have much explanation of why. Do you remember why you added it?

The problem @coskucinkilic is trying to solve is that if you click on an in-app link in a Popover, it doesn't result in an in-app navigation as with other in-app links but results in a complete page reload.

peterpeterparker:
@dskloetd as I said in my comment, I got the feeling that if you remove the stop propagation, it may results in the popover closing it self for example if the user clicks or select text in the popover. Question mark, so if it's not the case for sure it's fine.

dskloetd:
That would make sense if the backdrop were a parent of the element on which `stopPropagation` is removed. But it's not.

@coskucinkilic can you deploy a canister with this change to you DevEnv so we can all try and see if it causes any problems?

peterpeterparker:
> can you deploy a canister

Unrelated to the PR but to the comment: If the Juno DAO succeeds, one of the features hidden in the roadmap goal 'Better CI/GitHub integration' is the ability to easily deploy staging or other dev versions from PRs, similar to what's done on Vercel, for example.

coskucinkilic:
I have deployed changes in this PR and [this PR in nnsDapp](https://github.com/dfinity/nns-dapp/pull/5567) to devenv, feel free to review in[ this devenv](https://qsgjb-riaaa-aaaaa-aaaga-cai.cosku-ingress.devenv.dfinity.network/).

dskloetd:
Thanks. I've tested
1. Clicking on the popover does not close it.
2. Clicking on the copy button s in the popover does not close it.
3. Clicking on Manager Internet Identity opens Internet Identity in a separate tab and does not close the popover
4. Clicking Settings does close the popover and navigates to settings
5. Clicking Canisters closes the popover and navigates to canisters
This all seems correct.
@peterpeterparker what do you think?

peterpeterparker:
Sounds good. Thanks for the test, appreciated! Let's go then.

peterpeterparker (approved):
LGTM, thx

PR #492, opened 2024-10-03
NNS1-3377 update logo fix disappearance and make menu items scrollable

coskucinkilic:
# Motivation

Display new logo, make toggle and logo always visible, make menu items scrollable so it can be used in small devices.

# Changes

1 Logo on chain is updated
2 Bottom logo and toggle disappearance on small devices is removed.
3 Menu item inner section is now scrollable
4 Scrollbar is set to hidden because it was getting in between menu item background and page content when an item is selected.
5 Adjusted paddings
6 Removed unnecessary styling like sticky
7 Remove the div in Layout
8 Update Gix layout padding as we have the padding in Menu now.
9 Now role=menu width changes instead of inner and all transitions happen in menu except logos.
10 Update tests

# Screenshots


coskucinkilic:
this pr is closed and broken down into multiple prs, also implementation of scroll and navigation is different. You can check the main scroll pr [here.](https://github.com/dfinity/gix-components/pull/493)

PR #493, opened 2024-10-07
NNS1-3377 make inner scrollable

coskucinkilic:
# Motivation

Make Inner scrollable on small height devices we can display bottom logo without disturbing menu items.

# Changes

1. Add width and height properties to NNSlogo
2. Move nns logo and bottom logo from Menu background into Menu.
3. Add inner wrapper to add scroll around inner.
4. Update margins and paddings to match with old design as close as possible.
5. Update NNSLogo paddings in Layout
6. Add transitions for logos in Menu 
7.  Scrollbar is set to hidden because it was getting in between menu item background and page content when an item is selected.
8. Update gix layout paddings

# Screenshots

https://github.com/user-attachments/assets/047daa29-fad4-4840-abd0-ce20e2fdf2f5


dskloetd:
Can you provide some more context in the PR description? I already have trouble understanding it (what is Inner?) and in 3 months time it will be even harder to understand.

And why are there so many screenshot change? Making something scrollable should only create screenshot changes if there is a scrollbar in the new screenshot, no?

coskucinkilic:
Currently the menu inner part which hold the menu items is already scrollable you can go and test it on https://nns.ic0.app/ what i am trying to do here is to make it scrollable correctly.

What is wrong with the current setup, there is not enough padding at bottom and at top to scroll properly.
The logos are displayed in MenuBackground which we cannot use the logo heights to adjust the MenuItems and scroll nicely.
Also when we shrink the height of the browser logos are rendered on top of MenuItems.

Menu is closed on open based on two different states, MenuCollapsed and LayoutMenuOpen, and the close and open states change the width and left margin of the inner div. I think it is not a great practice to set margin to negative values to move elements but i kept using the same as this ticket already took a lot of my time.

What i have done, move logo and bottom logo from MenuBacktround into Menu.
Change it to flex div, created an inner wrapper around inner, which is the div that hold menu items, the link buttons and metrics and footer, to add scroll functionality. I have added some paddings in scrollablable area so we can scroll beyond menu items in both directions. Also to make the transitions funciton properly and not shrink NNSLogo height as we are shrinking the width i have added fix width and height to NNS logo.

As these padding changes and logos moving from background into Menu, there has been shifts in the paddings around, so i did what i can in gix-components and also create a separate pr in [NNSDapp](https://github.com/dfinity/nns-dapp/pull/5590), which is still in draft and waiting for this pr to be merged first. 






Hope it is a little bit more clear now, maybe i missed couple of things but this is the main gist of why there are so many changes.

dskloetd:
Sorry, I still don't understand why there need to be any screenshot changes.
If you need to add/remove padding/margin in one place, why can't you remove/add the same amount of padding/maring somewhere else, such that the total padding remains the same and elements remain in the same position on the screen?

coskucinkilic:
I am remove the paddings now, i think it is better but obviously it creating screenshots mismatch. 

Without the paddings i think scroll won't feel as nice as current version.

coskucinkilic:
closed in favor of [this pr](https://github.com/dfinity/gix-components/pull/498)

PR #498, opened 2024-10-09
Stop hiding bottom logo and menu collapse button

dskloetd:
# Motivation

We hide the bottom logo when the viewport is less than `654px` high. This is done to avoid the logo overlapping with the menu items.
This was necessary because the menu logos exist in a separate component, `MenuBackground` from `Menu` where the menu items are.

A better approach is to let the logos take up space next to the menu items. Then the bottom logo can just falls out of the viewport and a scrollbar can appear.

# Changes

1. Remove `MenuBackground`.
2. In `Menu` instead of reserving padding at the top for the logo in the background, place the logo in an element of the same size.
3. Put the bottom logo in the menu. The `flex-grow: 1` on `.slot-content` makes sure the bottom logo is at the bottom.
4. Stop hiding the menu collapse button.

Values are derived as follows:

1. The height of the element for the top logo is equal to the padding previously reserved for the top logo.
2. The top padding in the `.top-logo` element in small viewports (`--menu-stack`) is equal to the top margin in `.logo-nns` element in `MenuBackground`.
3. The top padding in the `.top-logo` element in large viewports (`--menu-stack + --padding`) is equal to the top margin plus the top padding in `.logo-nns` element in `MenuBackground`.
5. The bottom padding in the `.bottom-logo` element is equal to the bottom margin in the `.bottom-logo` element in `MenuBackground`.
6. The top padding in the `.bottom-logo` element is such that we can keep the same space between the bottom logo and the theme switcher in NNS dapp after removing the bottom margin in NNS dapp.

# Screenshots

Tested manually with NNS dapp at: https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/

image


coskucinkilic (approved):
All the changes in this PR looks lgtm, however I have one concern related to all inner scroll result.

We have implemented inner scroll mainly for users with screen with small heights however after all these changes, collapsed view of small screen rendered like this:

Screenshot 2024-10-16 at 16 58 28

dskloetd:
Good catch!
I found out that it also happens today so I think it's OK not to fix it this time.
But if we remove the TVL and buttons from the DOM when we collapse the menu, we can make the scrollbar less likely to appear when the menu is collapsed. Have a look: https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/

PR #500, opened 2024-10-14
Clean up menu padding

dskloetd:
# Motivation

We want to merge the `MenuBackground` into the `Menu` component and make the `Menu` scrollable.
To limit the complexity of that PR, I want to first clean up the existing CSS.

The menu width is [specified as 200px](https://github.com/dfinity/gix-components/blob/8c5367f60e971b74508e62faefe48bfd4da107c7/src/lib/styles/global/variables.scss#L5), but the actual menu panel is `8px` or `16px` wider because of an extra padding or margin depending on the width of the viewport.

To simplify removing `MenuBackground`, we move the padding into `.inner` so that it's easier to center the logos.
And we use the same mechanism for large and small screens instead of using a margin in one case and a padding in the other case.

# Changes

1. Split `padding` in the menu for large screens in `padding-top`, `padding-left`, `padding-right` and `padding-bottom`.
    1. Move `padding-left` into `.inner`.
    2. Removing `padding-right` and `padding-bottom` because they are 0.
2. Change `margin-left: var(--padding);` for small screens to `padding-left: var(--padding);` in `.inner`.
3. Add `var(--padding)` or `var(--padding-2x)` to the actual menu width to keep the same actual final menu width and make it explicit.
4. Apply the left padding inside a `.slot-content` element instead of on `.inner` to avoid `.inner` having a width when its width should be 0. This will also help latter to avoid padding on `.inner` causing the logos to be off center.

# Screenshots

The behavior and visuals should be unchanged and can be tested manually at https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/

### DIFF
--- a/src/lib/components/Menu.svelte
+++ b/src/lib/components/Menu.svelte
@@ -27,7 +27,9 @@
     on:click={close}
     on:keypress={($event) => handleKeyPress({ $event, callback: close })}
   >
-    <slot />
+    <div class="slot-content">
+      <slot />
+    </div>
   </div>
 
   <button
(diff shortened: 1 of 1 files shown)


src/lib/components/Menu.svelte
```diff
@@ -61,25 +63,35 @@
...
     // otherwise the first selected menu entry would be cut off in mobile view.
     padding-top: var(--menu-selection-outer-radius);
 
+    .slot-content {
+      display: flex;
+      flex-direction: column;
+
+      gap: var(--padding-0_5x);
+
+      padding-left: var(--padding);
```
coskucinkilic, line 88: As you are cleaning maybe it would be ideal to store this paddings at top as a variable also like 91 and use them in line 120 and line 113 .

Then we will be building on single variable for padding left.

Not necessary but would be nice

coskucinkilic (approved):

PR #503, opened 2024-10-16
test: Add testId to Popover

AntonioVentilii-DFINITY:
# Motivation

Useful to have a `testId` in the `Popover`, for example, to check if a menu opened but the elements inside may be not mounted.


src/lib/components/Popover.svelte
```diff
@@ -11,6 +11,7 @@
   export let direction: "ltr" | "rtl" = "ltr";
   export let closeButton = false;
   export let invisibleBackdrop = false;
+  export let testId: string | undefined = undefined;
```
dskloetd, line 14: How about defaulting to `"popover-component"`

dskloetd (approved):

src/lib/components/Popover.svelte
AntonioVentilii-DFINITY: Yes, better! Changed, thank you for the suggestion

dskloetd (approved):

PR #505, opened 2024-10-16
Disable triggering events if checkbox is disabled

coskucinkilic:
# Motivation

 Checkbox currently triggers keypress event on keypress and nnsChange event on click, eventhought it is disabled. Disabled input elements shouldn't trigger these events, if we need information related to disabled Checkbox events, those should be implemented separately.

# Changes

Update Checkbox to not trigger events if disabled.

# Screenshots

no change

### DIFF
--- a/src/lib/components/Checkbox.svelte
+++ b/src/lib/components/Checkbox.svelte
@@ -15,11 +15,13 @@
    * Emit an event when the checkbox or container is clicked. The state should be updated by consumer.
    */
   const onClick = ($event: MouseEvent | TouchEvent) => {
-    if (preventDefault) {
-      $event.preventDefault();
-    }
+    if (!disabled) {
+      if (preventDefault) {
+        $event.preventDefault();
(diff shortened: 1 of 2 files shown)


src/lib/components/Checkbox.svelte
```diff
@@ -28,7 +30,8 @@
 <div
   tabindex="0"
   on:click|preventDefault={onClick}
-  on:keypress={($event) => handleKeyPress({ $event, callback: change })}
+  on:keypress={($event) =>
+    !disabled && handleKeyPress({ $event, callback: change })}
```
dskloetd, line 34: I was wondering how a disabled checkbox would even get a keypress event because how can you focus a disabled checkbox?
So then I thought maybe the `tabindex` should be removed if the checkbox is disabled?

But feel free to consider this outside the scope of the PR.

dskloetd (approved):

PR #506, opened 2024-10-16
Organise Tag css variables

mstrasinskis:
# Motivation

Currently, neither the CSS variable names nor the values are in sync with Design for the Tag component.

# Changes

- Rename `--text-text` and `--text-light` to be more Tag related (they are used in the Tag component only).
- Add separate variables for the default Tag color and background to match the structure in Figma. This changes the default Tag colors, so they now match the design.

# Screenshots

## Tag default colour changes

| Before | After |
|--------|--------|
| image | image |
| image | image |


src/lib/components/Tag.svelte
```diff
@@ -15,22 +15,22 @@
...
     &.success {
       background-color: var(--positive-emphasis);
-      color: var(--text-light);
+      color: var(--tag-text-light);
     }
 
     &.warning {
       background-color: var(--warning-emphasis);
-      color: var(--text-text);
+      color: var(--tag-text-dark);
```
dskloetd, line 28: Where do we use the `warning` and `error` tags and what do they look like?

dskloetd (approved):

src/lib/components/Tag.svelte
mstrasinskis: I don’t think those styles are used in the NNS dapp, so they’re here for backward compatibility.

image

PR #514, opened 2024-10-25
Replace same variables

mstrasinskis:
# Motivation

Currently, we have primitive colors with the same values. In this update, we consolidate them into a single value and replace references in the themes with non-deprecated ones that share the same value. None of them are used in the nns-dapp repo.

# Changes

- Update CSS variables.

### DIFF
--- a/src/lib/styles/global/colors.scss
+++ b/src/lib/styles/global/colors.scss
@@ -11,26 +11,27 @@
   --cp-light-100: var(--cp-light-50);
   --cp-light-125: #dadef2;
   // @deprecated (same as --cp-light-125)
-  --cp-light-150: #dadef2;
+  --cp-light-150: var(--cp-light-125);
   --cp-light-200: #eef1fe;
   --cp-light-250: #6f7fcd;
   // @deprecated
   --cp-light-400: #34407c;
   // @deprecated (same as --cp-light-600)
-  --cp-light-500: #3d4d99;
+  --cp-light-500: var(--cp-light-600);
   --cp-light-600: #3d4d99;
   --cp-light-900: #151a33;
-  // @deprecated
+  // @deprecated TODO: replace with Neutral-50A45
   --cp-light-opaque: #ffffff40; // --cp-light-50 with 40% opacity
   --cp-light-gradient-1: #d4e7fa;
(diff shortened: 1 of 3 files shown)


src/lib/styles/global/colors.scss
```diff
@@ -11,26 +11,27 @@
...
   --cp-light-250: #6f7fcd;
   // @deprecated
   --cp-light-400: #34407c;
   // @deprecated (same as --cp-light-600)
-  --cp-light-500: #3d4d99;
+  --cp-light-500: var(--cp-light-600);
   --cp-light-600: #3d4d99;
   --cp-light-900: #151a33;
-  // @deprecated
+  // @deprecated TODO: replace with Neutral-50A45
```
dskloetd, line 23: What is `Neutral-50A45`?

mstrasinskis: A new name for the `--cp-light-hover-overlay` variable.

dskloetd: But I don't see where it's defined.
Also for `--cp-light-hover-overlay` it says `same as --cp-light-50-45` so should that also say to use Neutral-50A45 instead?
And why such a strange name? It seems like a combination between a name and a color code. Shouldn't it be either/or?
What does 50A45 stand for?
And can we stick to lower kebab case for these variables?

mstrasinskis: It’s a CUSTOM adaptation of the Material Design approach by our designers, where the numbers represent the shade or intensity of the color. The `A` stands for alpha.

dskloetd: No, we don't use underscores anywhere in this file. And neither do we use upper case.
It's all a combination of lower case and hyphens. Like `--cp-light-hover-overlay`.
Unless you plan to switch everything to title case, why would we diverge from what we currently do?

And again, can you tell me where it's defined in gix-components?

mstrasinskis: Got it, sorry for the misunderstanding. I pasted the color name from Figma as a note to myself to remember to replace it. The name in the front-end will be `--neutral-50A45`.

dskloetd (approved):
I still don't get it but I guess you're removing it anyway in the next PR.

PR #516, opened 2024-10-28
Rename css colours

mstrasinskis:
# Motivation

nns-dapp and Figma use different names for the primitive colors. Here, we rename them to align with the design variables.

Note: Some primitives are used directly in nns-dapp. These variables were left with a TODO to rename them later.

# Changes

- Rename CSS variable in the colors.scss and theme files.

# Tests

Tested locally with nns-dapp end-to-end screenshot tests against the upgraded gix-components.
And manually, no changes noticed.


src/lib/styles/global/colors.scss
```diff
@@ -81,4 +81,67 @@
   --green-tint: #2db28625;
   --pink: #f2556f;
   --pink-tint: #f2556f25;
+
```
dskloetd, line 84: Can you add a TODO here?
```
// TODO: Remove all variables above this line once they are all replaced with new variables.
```

dskloetd (approved):

PR #520, opened 2024-10-30
feat: shrink Popover if overflows

AntonioVentilii-DFINITY:
# Motivation

We would like to contain the `Popover` in the visible viewport, making the content scrollable if otherwise.

# Changes

- Wrap the `Popover` content in a `div` element: we don't want to make the close button part of the scrollable part.
- Make the vertical overflow `auto` for the new container.
- Set the `max-height` of the `Popover` to be contained in the viewport minus a custom amount of padding (I put 6, but we can discuss it).

# Screenshots

I included one more E2E test and a rendering test, but here a visual video.

https://github.com/user-attachments/assets/5e5856f4-b72e-4772-aead-3c10bf42439b


src/lib/components/Popover.svelte
```diff
@@ -88,6 +90,8 @@
     // limited by `100vw - right padding`
     max-width: calc(100vw - var(--padding));
 
+    max-height: calc(100vh - var(--popover-top) - calc(6 * var(--padding)));
```
peterpeterparker, line 93: Have you test on iOS?

peterpeterparker (commented):
Looks good, just wondering if we should have a test on iOS before approving and merging.

src/lib/components/Popover.svelte
AntonioVentilii-DFINITY: I don't why, my local iOS simulator is not starting with Safari... so I just turned on the responsive mode in Safari:


https://github.com/user-attachments/assets/231e6300-ae87-4470-b30d-cbe5bb10e6f1

peterpeterparker: It seems to work partially. See screenshot. `100vh` is always tricky on iOS.

Capture d’écran 2024-11-01 à 07 11 15

peterpeterparker: I guess it's related the url bar being at the bottom and not accounted with `vh`.

For iOS maybe `svh` instead of `vh`? or `dvh`?

Capture d’écran 2024-11-01 à 07 14 11

AntonioVentilii-DFINITY: I went with `dvh`, WDYT?


Uploading Screen Recording 2024-11-01 at 13.59.27.mov…

peterpeterparker: Nice!

peterpeterparker: Cool, did not knew that worked 👍

You maybe don't need the second if, like vh just default? Also safer for those old phone which don't support `@supports`?

```
@supports not (height: 100dvh) {
```

AntonioVentilii-DFINITY: makes sense! i removed it and make the calculation fallback to `100vh`

peterpeterparker (approved):
LGTM, thx

PR #523, opened 2024-10-30
Analyze theme variables

dskloetd:
I analyzed the `dark.scss` and `light.scss` files to get an idea of what theme variables we might need.
A theme variable will have a value for dark and a different value for light and can be used by multiple components.
To understand which component variables might be able to use the same theme variable we need to know which component variables have the same value in both the light and the dark theme.
Additionally, variables that exist in only one of the themes or that have the same value in both themes, might not need a theme variable.

I changed the input files to make sure that each key/value pair is on a single line to make parsing easier.

Then I ran `node theme.js > output.text` and added both to the branch.

### DIFF
--- a/output.text
+++ b/output.text
@@ -1,0 +1,346 @@
+Reading file: src/lib/styles/themes/dark.scss
+The following lines are ignored:
+/** Default theme: dark */
+:root {
+/* Light/Text/Text-xx? */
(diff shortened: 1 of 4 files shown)


AntonioVentilii:

Thanks for the contribution!

PR #518, opened 2024-10-30
tmp: Theme level variables

mstrasinskis:
# Motivation

DO NOT MERGE!

## Existing in old dark but not in new dark: Map(7) {
  'box-shadow-color' => 'night-500',
  'box-shadow-color-rgb' => '43, 51, 67',
  'dropdown-focus-background' => 'night-575',
  'dropdown-focus-border-color' => 'blue-accent',
  'focus-background-tint' => 'night-400',
  'table-background' => 'night-500',
  'tag-text-dark' => 'night-550'
}
## Existing in old light but not in new light: Map(7) {
  'box-shadow-color' => 'night-500',
  'box-shadow-color-rgb' => '111, 89, 133',
  'dropdown-focus-background' => 'neutral-75',
  'dropdown-focus-border-color' => 'blue-accent',
  'focus-background-tint' => 'neutral-50',
  'table-background' => 'neutral-75',
  'tag-text-dark' => 'blue-900'
}
#### Distance > 50%
  tooltip-divider diff_rgb:73.52% diff_alpha:91.76% (from 3d3061 $${\color{#3d3061}⏺}$$ to ffffff15 $${\color{#ffffff}⏺}$$) [usage](https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/dfinity/%28gix-components%7Cnns-dapp%29%24+tooltip-divider&patternType=keyword&sm=0)
  tooltip-border-color diff_rgb:100.00% diff_alpha:14.51% (from 00000020 $${\color{#000000}⏺}$$ to ffffff45 $${\color{#ffffff}⏺}$$) [usage](https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/dfinity/%28gix-components%7Cnns-dapp%29%24+tooltip-border-color&patternType=keyword&sm=0)
  backdrop-contrast diff_rgb:87.34% diff_alpha:0.00% (from ffffff $${\color{#ffffff}⏺}$$ to 151a33 $${\color{#151a33}⏺}$$) [usage](https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/dfinity/%28gix-components%7Cnns-dapp%29%24+backdrop-contrast&patternType=keyword&sm=0)
[...]

PR #524, opened 2024-10-31
Replace dark theme

mstrasinskis:
# Motivation

Update the dark theme.

Demo: https://qsgjb-riaaa-aaaaa-aaaga-cai.mstr-ingress.devenv.dfinity.network/ (+[blue icons update](https://github.com/dfinity/gix-components/pull/519))

# Changes

1. Added new color that is needed for the new theme.
2. Copied CSS variables from the dark theme to the light theme to address any missing variables in the light theme. To not introduce any light theme changes after color replacement in the dark theme.
3. Replaced colors in the dark theme with new values. Color mapping was done using another [branch](https://github.com/dfinity/gix-components/tree/theme-level-variables) as a color source, where designers verified the new new theme colors. Using [this script](https://github.com/dfinity/gix-components/pull/518/files#diff-1388fad36a3dfb5bf5ab3d3635ffed1a4772144be7b7eeeab59d1eb0673b3ef1) and manual adjustment not covered by the script entries (mainly hex-to-RGB conversions).

# Screenshots

image
image


src/lib/styles/global/colors.scss
```diff
@@ -121,6 +121,7 @@
   --violet-650: #150c31;
   --violet-900: #130a2c;
 
+  --night-50: rgb(155, 173, 209);
```
dskloetd, line 124: Why is this `rgb` while all the others are `#XXXXXX`?

src/lib/styles/themes/light.scss
```diff
@@ -138,7 +168,6 @@
 
     /* Light/Text/Text */
     --text-color: var(--blue-900);
-    // TODO-colors: Remove application of rgbs
```
dskloetd, line 141: Why did you removed these TODOs? I don't see anything about it in the PR descriptions.

src/lib/styles/global/colors.scss
mstrasinskis: Good catch, this should be hex.

src/lib/styles/themes/light.scss
mstrasinskis: I initially removed them because the comments were inconsistent, and this was already addressed in another branch. But you’re right; removing them here is confusing. Restored.

src/lib/styles/themes/dark.scss
```diff
@@ -161,56 +163,56 @@
...
-  --card-background-shade: var(--violet-500);
-  --card-background-tint: var(--violet-400);
-  --card-background-disabled: var(--cp-dark-1000-20);
+  --card-background: var(--night-500);
+  --card-background-rgb: 43, 51, 67;
+  --card-background-contrast: var(--neutral-50);
+  --card-background-contrast-rgb: 255, 255, 255;
+  --card-background-shade: var(--night-575);
+  --card-background-tint: var(--night-400);
+  --card-background-disabled: var(--neutral-blue-a10);
```
dskloetd, line 185: I noticed very low contrast on the neuron badges (on the left) and disabled buttons (on the right).
Please work with Artem on these before merging.
image

dskloetd (approved):

src/lib/styles/themes/dark.scss
mstrasinskis: Thanks, updated
image

dskloetd: I mentioned 2 things and it looks like you changed 7 things.
What are the other 5?

mstrasinskis: Requests from Artem. I committed them separately, to keep it clear.
1. "Increase icons on card contrast" (you mentioned)
2. "Swap modal backgrounds" to be same as in Figma and avoid having contrast issue after the next step (3.).
3. "Increase disabled button contrast" (you mentioned).
4. "Increase scrollbar contrast" Artem mentioned.
5. "Make the disabled input state more distinguishable" Artem mentioned.

PR #526, opened 2024-11-08
NNS1-3435: Updates thumbs-up and thumbs-down icons

yhabib:
# Motivation

Replace original google font icons with a Dfinity version.

# Changes

- Replaces svgs 
- Adds size prop to make them responsive.

# Screenshots

Screenshot 2024-11-08 at 16 38 15

### DIFF
--- a/src/lib/icons/IconThumbDown.svelte
+++ b/src/lib/icons/IconThumbDown.svelte
@@ -1,12 +1,26 @@
-<!-- source: https://fonts.google.com/icons?selected=Material%20Icons%3Athumb_down%3A -->
+<!-- source: DFINITY foundation -->
+<script lang="ts">
+  import { DEFAULT_ICON_SIZE } from "$lib/constants/constants";
+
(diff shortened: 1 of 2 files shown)


yhabib:
Note: There was a misunderstanding regarding the design, and we no longer need the circular version of these icons. I will remove them in a future PR.

src/lib/icons/IconThumbDown.svelte
```diff
@@ -1,12 +1,26 @@
...
+
+  export let size = `${DEFAULT_ICON_SIZE}px`;
+</script>
+
 <svg
   xmlns="http://www.w3.org/2000/svg"
-  height="18px"
-  viewBox="0 0 24 24"
-  width="18px"
+  width={size}
```
dskloetd, line 10: The previous version was `18px` and the new one is `20px` by default.
Does this look good in NNS dapp in places where we don't specify the size?
Or will we have to apply a size of `18px` in those places when we update the gix-components dep?

dskloetd (approved):

PR #528, opened 2024-11-12
NNS1-3435: adds a new semantic color variable to the themes

yhabib:
# Motivation

A new theme requires a new light variable to be used for negative emphasis connotation.

# Changes

-  A new semantic color variable has been added to the themes.


mstrasinskis (approved):
Thanks!

PR #529, opened 2024-11-12
Tag color variables

mstrasinskis:
# Motivation

Currently, primitive colors are used in the proposal status tags. To make it more organized, we are creating separate variables for each proposal status (except for the status Unknown, as it’s not a real status and uses a ~~disabled~~ default style).

# Changes

- Added new variables that match the design.

# Screenshots

No visual changes yet.


src/lib/styles/themes/dark.scss
```diff
@@ -218,4 +218,15 @@
   --negative-emphasis: var(--pink-200);
   --negative-emphasis-contrast: var(--neutral-50);
   --negative-emphasis-light: var(--pink-200-a25);
+
+  --proposal-open-background: var(--blue-accent-tint);
```
yhabib, line 222: do I understand correctly that our pattern is something like `component-state-property`?

mstrasinskis: Good point! I found them in the new design variables in Figma. I’ll rename them accordingly.
image

src/lib/styles/themes/dark.scss
```diff
@@ -218,4 +218,18 @@
   --negative-emphasis: var(--pink-200);
   --negative-emphasis-contrast: var(--neutral-50);
   --negative-emphasis-light: var(--pink-200-a25);
+
+  --elements-tag-background: var(--night-575);
```
dskloetd, line 222: I'm wondering if we really need the "elements" prefix here. Do you know what purpose it serves in Figma?

mstrasinskis: Good point; let’s keep it simple and flatter. After syncing with Artem, we decided to remove the elements. Updated.

dskloetd (approved):

PR #531, opened 2024-11-25
feat: add test id to Toggle

AntonioVentilii:
# Motivation

Useful to have a test ID for the toggle.

### DIFF
--- a/src/lib/components/Toggle.svelte
+++ b/src/lib/components/Toggle.svelte
@@ -7,13 +7,14 @@
   export let checked: boolean;
   export let ariaLabel: string;
   export let disabled = false;
+  export let testId = 'toggle';
 
   const dispatch = createEventDispatcher();
 
   const id = nextElementId("toggle-");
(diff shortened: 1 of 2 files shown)


src/lib/components/Toggle.svelte
```diff
@@ -7,13 +7,14 @@
...
   export let disabled = false;
+  export let testId = "toggle";
 
   const dispatch = createEventDispatcher();
 
   const id = nextElementId("toggle-");
 </script>
 
-<div class="toggle" class:disabled>
+<div class="toggle" class:disabled data-tid={testId}>
```
peterpeterparker, line 17: Maybe update or add a test that uses it in `Toggle.spec.ts`?

AntonioVentilii: Done!

peterpeterparker: Why not

image

PR #533, opened 2024-11-26
Use same color for the backdrop

mstrasinskis:
# Motivation

To sync with the latest [design](https://www.figma.com/design/29OdZHbNaOk9RkrfESucjb/NNS-1.03-WIP?node-id=10663-13919&t=3Bc5ARspbSXrBliO-4), we are introducing a new primitive color and applying it to both themes.

# Changes

- Add new color.
- Use it for backdrops.

# Screenshots

| Before | After |
|--------|--------|
| image | image |
| image | image |


src/lib/styles/global/colors.scss
```diff
@@ -106,6 +106,7 @@
   --neutral-blue-a45: #ccddff45;
   --neutral-blue-800-a15: #33468015;
   --neutral-blue-800-a20: #33468020;
+  --neutral-blue-1000-a65: #05070aa6;
```
yhabib, line 109: I only see this one in the Figma design. 🤔

Screenshot 2024-11-26 at 17 28 57

mstrasinskis: We haven’t migrated to the new colors yet, but this is a good point. I’ve asked to add the missing primitive.

mstrasinskis: You can find a new primitive here  https://www.figma.com/design/R0OMkHgUeHJFsK3woAZl4D/NNS-Library-2.0?node-id=3967-2941&node-type=canvas&t=KanbpnOxMH3PUugr-0

yhabib: I still can't find it but I trust you :) 

Screenshot 2024-11-27 at 09 20 28

yhabib (approved):

src/lib/styles/global/colors.scss
mstrasinskis: The new primitive was added here
image
But you're right, the backdrop hasn't been updated yet, 'll ping the designer.

PR #539, opened 2024-12-05
style: force content-box for Checkbox

AntonioVentilii:
# Motivation

It is useful to "force" the Checkbox component to have the CSS property `box-sizing: content-box;`. That is because, if for instance it is used in a context where the property is set as border-box, the results are not so pretty:

Screenshot 2024-12-05 at 19 38 11

While, after this change, we have the intended results in all type of contexts:

Screenshot 2024-12-05 at 19 38 59


peterpeterparker (approved):
LGTM, thx

PR #540, opened 2024-12-07
New icon IconCheckCircleFill

mstrasinskis:
# Motivation

A new icon is needed to display the neuron’s active rewards state.

# Changes

- New IconCheckCircleFill.

# Screenshots

image


src/lib/icons/IconCheckCircleFill.svelte
```diff
@@ -0,0 +1,18 @@
+<!-- source: DFINITY foundation -->
+<script lang="ts">
+  export let size = `24px`;
```
yhabib, line 3: is this intended? shouldn't it be `DEFAULT_ICON_SIZE`?

yhabib (approved):

src/lib/icons/IconCheckCircleFill.svelte
mstrasinskis: Good catch! I blindly copied it from IconCheckCircle.svelte. It’s now updated.

PR #541, opened 2024-12-09
build: vite v6

peterpeterparker:
# Motivation

Svelte v5 and notably `@sveltejs/vite-plugin-svelte` uses Vite v6 in their last version. Therefore we also want to keep the pace and use those version.


peterpeterparker:
Let's use this PR to migrate to vite v6. Currently blocked because of an incompatibility in Vitest v3. 

See https://github.com/sveltejs/vite-plugin-svelte/issues/1043#issuecomment-2529175816

peterpeterparker:
We gonna bump vite and vitest at the same time as Svelte otherwise we would get warnings because of inter dependencies constain.

PR #548, opened 2024-12-10
build: svelte v5 (without components migration)

peterpeterparker:
# Motivation

We want to use the most recent tools, but we also want to avoid being left behind, notably in terms of security. For example, SvelteKit is going to patch a minor issue with cookies only in a future version. That's why we are migrating to Svelte v5.

# Changes

This PR updates the dependencies and makes the required configuration changes. No components are migrated since Svelte v5 is backward compatible with how Svelte v4 is composed.


package.json
```diff
@@ -86,7 +87,7 @@
   ],
   "peerDependencies": {
     "@dfinity/utils": "*",
-    "svelte": "^4.2.1"
+    "svelte": "^5"
```
mstrasinskis, line 90: Is only the major version specified on purpose?

PR #554, opened 2024-12-17
test: render with events utility refactoring

peterpeterparker:
# Motivation

Using an `$on` callback for testing is deprecated in Svelte v5 (PR #548). Instead, we will now use the `events` option of `render`, which itself will eventually be replaced by the library. To minimize changes in the Svelte v5 PR, we refactored the test to use a utility function that handles rendering and event binding.

# Changes

- Introduce and use `renderWithEvents` utilitiy.

### DIFF
--- a/src/tests/utils/render.test-utils.ts
+++ b/src/tests/utils/render.test-utils.ts
@@ -1,0 +1,20 @@
+import { render, type RenderResult } from "@testing-library/svelte";
+import type { ComponentProps, ComponentType, SvelteComponent } from "svelte";
+
+export const renderWithEvents = <C extends SvelteComponent>(
+  cmp: ComponentType<C>,
+  options: {
+    props?: ComponentProps<C>;
+    events: Record<string, ($event: CustomEvent) => void>;
+  },
+): RenderResult<C> => {
+  const { component, ...rest } = render(cmp, { props: options.props });
+
+  const events = Object.entries(options.events);
+
+  events.forEach(([event, fn]) => {
+    component.$on(event, fn);
+  });
+
+  return { component, ...rest };
+};
--- a/src/tests/lib/components/Back.spec.ts
+++ b/src/tests/lib/components/Back.spec.ts
@@ -1,13 +1,14 @@
 import Back from "$lib/components/Back.svelte";
-import { fireEvent, render } from "@testing-library/svelte";
+import { fireEvent } from "@testing-library/svelte";
+import { renderWithEvents } from "../../utils/render.test-utils";
 
 describe("Back", () => {
   it("should forward the click event", () =>
     new Promise<void>((done) => {
-      const { getByTestId, component } = render(Back);
-
-      component.$on("nnsBack", () => {
-        done();
+      const { getByTestId } = renderWithEvents(Back, {
+        events: {
+          nnsBack: () => done(),
+        },
       });
 
       const button = getByTestId("back") as HTMLButtonElement;
--- a/src/tests/lib/components/Card.spec.ts
+++ b/src/tests/lib/components/Card.spec.ts
@@ -1,5 +1,6 @@
 import Card from "$lib/components/Card.svelte";
 import { fireEvent, render } from "@testing-library/svelte";
+import { renderWithEvents } from "../../utils/render.test-utils";
 
 describe("Card", () => {
   it("should render an article", () => {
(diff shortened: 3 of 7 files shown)


src/tests/utils/render.test-utils.ts
```diff
@@ -0,0 +1,20 @@
+import { render, type RenderResult } from "@testing-library/svelte";
+import type { ComponentProps, ComponentType, SvelteComponent } from "svelte";
+
+export const renderWithEvents = <C extends SvelteComponent>(
```
dskloetd, line 4: FYI, in nns-dapp we already have a custom `render` function:
https://github.com/dfinity/nns-dapp/blob/73490d2e89f8b73b33463552b9bfd692d0dd3433/frontend/src/tests/utils/svelte.test-utils.ts#L5

So having a second custom `render` function might not work and it probably makes more sense there to include both functionality in the same `render` function.

With that in mind, I think it might also make sense to just call this function `render` instead of `renderWithEvents` and keep it as a catch-all for things we may have to do to deal with quirks of the underlying `render` function. Perhaps we could even make all calls to `render` go through this (in another PR) so that in the future making adjustments is even easier.

No strong opinion but something to consider.

src/tests/utils/render.test-utils.ts
```diff
@@ -0,0 +1,20 @@
...
+  cmp: ComponentType<C>,
+  options: {
+    props?: ComponentProps<C>;
```
dskloetd, line 7: How about already including `events` in `props` so we offer the same API as the Svelte 5 render function?

dskloetd (approved):

src/tests/utils/render.test-utils.ts
peterpeterparker: > So having a second custom render function might not work 

I'm not sure what you mean. Both are currently unrelated and exist in different projects. Of course, if we notice during the Svelte v5 migration that such a utility is needed in the NNS dapp codebase, we might consider reusing the same function. However, even then, I'm not sure the NNS dapp should rely on a function from Gix Cmp for testing purposes, kind of feel it's two separate concerns.

> keep it as a catch-all for things 

I feel like the proposed approach introduces fewer changes, which makes the migration easier to achieve. I'm also unsure how the use of events will evolve in the future.

But if you would like to give it a try, not against for sure!

src/tests/utils/render.test-utils.ts
peterpeterparker: In Svelte v5 the `events` are not included in the `props` but within the `options` as in this API.

For example.

```
const { container } = render(Checkbox, {
        props,
        events: {
          nnsChange: () => done(),
        },
      });
```

src/tests/utils/render.test-utils.ts
dskloetd: Sorry for the long rambling. The only thing I was proposing now is to name the function just `render` instead of `renderWithEvents`. The rest was just trying to explain why I think it makes sense. But if it doesn't make sense to you, it's not important.

src/tests/utils/render.test-utils.ts
dskloetd: Oh, ok, great.
But then what do you mean by the in the other file:
```
// TODO: remove once events is migrated to props
```

src/tests/utils/render.test-utils.ts
peterpeterparker: Ah! Sure!

peterpeterparker: Done

src/tests/utils/render.test-utils.ts
peterpeterparker: I can probably rephrase "props" to "callback props" to be more precise.

In Svelte v5 events with dispatcher are deprecated and have to be replaced with callback props. Once we have migrated the events, we can pass callback as props instead of listening to events.

https://svelte.dev/docs/svelte/v5-migration-guide#Event-changes-Component-events

dskloetd: So the `events` that we are passing now are at the same time new and already deprecated?

Is that TODO specific to that one test or would it apply to all tests? I don't see what's different about that one test with the TODO.
Anyway, the PR is fine so feel free to merge and my questions can be answered by future PRs.

dskloetd (approved):

peterpeterparker:
> So the events that we are passing now are at the same time new and already deprecated?

Yes, kind of. We do it this way so we don’t have to migrate all components at once. It reduces the number of changes needed for the upgrade by solely adapting few tests.

> Is that TODO specific to that one test or would it apply to all tests?

Once this PR is merged and rebased into the Svelte v5 branch, only one TODO will remain.

PR #557, opened 2024-12-23
Add banner bg variable

mstrasinskis:
# Motivation

To solve the low contrast of the banners in cards and standardize their appearance, a new color theme variable has been added.

# Changes

- Added `--banner-background` to both themes.

# Screenshots

| Before | After |
|--------|--------|
| image | image |
| image | image | 
| image | image | 
| image | image |

### DIFF
--- a/src/lib/styles/themes/dark.scss
+++ b/src/lib/styles/themes/dark.scss
@@ -231,4 +231,6 @@
   --tag-executed-background: var(--orchid-200-a25);
   --tag-failed-text: var(--orange-200);
   --tag-failed-background: var(--orange-200-a25);
+
+  --banner-background: var(--card-background-tint);
 }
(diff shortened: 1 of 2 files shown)


yhabib (approved):
Thanks.
We can use it [here](https://github.com/dfinity/nns-dapp/blob/4b5c9c3f2a5f4f913c708fc17ec1e24c6b813e97/frontend/src/lib/components/ui/UsdValueBanner.svelte#L105)

mstrasinskis:
> Thanks. We can use it [here](https://github.com/dfinity/nns-dapp/blob/4b5c9c3f2a5f4f913c708fc17ec1e24c6b813e97/frontend/src/lib/components/ui/UsdValueBanner.svelte#L105)

Correct. And in the other custom implementations.

PR #558, opened 2025-01-09
Avoid margin when no title

mstrasinskis:
# Motivation

On the neuron details page, the voting power block needs to be displayed within a collapsible section. To achieve this, we can’t use the section’s title slot and must instead rely on a custom title (within the KeyValuePairInfo) placed in the section’s description block. Currently, if the title slot is left empty, an extra margin is added, which is undesirable.

image

# Changes

- Do not add the margin, when no title provided.

- Tested locally against nns-dapp.

# Screenshots

image

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -131,7 +131,7 @@ jobs:
         run: npm ci
       - name: Run Playwright tests
         run: npm run e2e:ci
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         if: failure()
         with:
           name: playwright-report
--- a/src/lib/components/Section.svelte
+++ b/src/lib/components/Section.svelte
@@ -3,10 +3,12 @@
 
   let noDescription = false;
   $: noDescription = $$slots.description === undefined;
+  let noTitle = false;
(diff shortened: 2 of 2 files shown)


.github/workflows/checks.yml
```diff
@@ -131,7 +131,7 @@ jobs:
         run: npm ci
       - name: Run Playwright tests
         run: npm run e2e:ci
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
```
yhabib, line 134: is this change related?

src/lib/components/Section.svelte
```diff
@@ -3,10 +3,12 @@
 
   let noDescription = false;
   $: noDescription = $$slots.description === undefined;
+  let noTitle = false;
+  $: noTitle = $$slots.title === undefined && $$slots.end === undefined;
```
yhabib, line 7: is `end` related to the title?

.github/workflows/checks.yml
mstrasinskis: No, it was used to test this fix https://github.com/dfinity/gix-components/pull/559. And will be reset after the fix is merged.

yhabib (approved):

PR #560, opened 2025-01-12
feat: remove analytics

peterpeterparker:
# Motivation

We do not really use the analytics for this project, therefore I think we can remove those. Plus, once removed, I'll be able to move the Satellite from my own account on Juno to one that we share with few other colleagues at the foundation.

# Changes

- Remove dependency, code and config for Juno Analytics


yhabib (approved):
Thanks

Arsamm1392:
> Thanks

PR #563, opened 2025-01-20
chore: disable ssr for showcase app

peterpeterparker:
# Motivation

Without disabling SSR we would not be able to build the showcase case app once the component `InfiniteScroll` is effectively used in the UI. This is because `IntersectionObserver` is not a known NodeJS property.

# Notes

PR that uses the `InfiniteScroll` in the UI is #562.

# Changes

- Disable SSR which we do not used in first place anyway
- Added wait in the segments E2E as the screenshots seem to be generated earlier following this change (see test https://github.com/dfinity/gix-components/actions/runs/12868805408/job/35876230527)


e2e/segment.e2e.ts
```diff
@@ -19,6 +19,9 @@ test("First segment is selected", async ({ page }) => {
   await showcase.scrollIntoViewIfNeeded();
   await clickToSegment(page, 0);
 
+  // Wait for animation
+  await page.waitForTimeout(750);
```
dskloetd, line 23: Is it possible to wait for a specific element to be present?
I think it would be more robust and faster.

e2e/segment.e2e.ts
```diff
@@ -19,6 +19,9 @@ test("First segment is selected", async ({ page }) => {
   await showcase.scrollIntoViewIfNeeded();
   await clickToSegment(page, 0);
 
+  // Wait for animation
```
dskloetd, line 22: Is it really about animation? Does disabling server side rendering cause animations to happen that otherwise wouldn't have happened?

peterpeterparker:
@dskloetd did not had time to turn it back to draft, still debugging. I start thinking there is actually an issue in the implementation which become unmasked, somehow, by this change but, as I said, still debugging. Turned it into a draft.

e2e/segment.e2e.ts
peterpeterparker: Change ultimatelly reverted.

e2e/segment.e2e.ts
peterpeterparker: Change ultimatelly reverted.

peterpeterparker:
Display issue - or rendering race condition - was indeed reproducible on main and improved in ##565.

AntonioVentilii (approved):
LGTMtks

PR #564, opened 2025-01-20
chore: e2e local development

peterpeterparker:
# Motivation

It's easier to develop E2E tests while being able to run those against a local server and not in headless mode. This PR provides such a setup.

# Changes

- Add new `npm run e2e:dev` command line
- If dev mode, then use dev server and no headles in the Playwright config


playwright.config.ts
```diff
@@ -1,16 +1,27 @@
...
-    command: "npm run staging && npm run preview",
-    port: 4173,
-  },
+  webServer,
   testDir: "e2e",
   testMatch: ["**/*.e2e.ts"],
   use: {
     testIdAttribute: "data-tid",
     trace: "on",
+    ...(DEV && { headless: false }),
```
dskloetd, line 24: Do we need to set this in the default config? You can pass `--headed` when running `playwright` if you don't want headless.

dskloetd (approved):

playwright.config.ts
peterpeterparker: I personally prefer to have the tests running visually all the time when developing locally—it's my default workflow. 
However, I don't have a strong opinion on it. If you prefer to remove the default, I can do so. Let me know.

dskloetd: I'm used to adding `--headed` when I want and to not add it when I don't want.
Is it even still possible to run not headed with this config?

peterpeterparker: Removed.

PR #565, opened 2025-01-21
feat: observe resize segment for better indicator display

peterpeterparker:
# Motivation

Tests in #563 were failing. After debugging, I figured out that the potential display issue actually already existed on the main branch. Specifically, since the segment buttons have no fixed width, depending on the rendering, it's possible that on mount, the buttons are not yet fully rendered—meaning they haven't occupied their full width yet—resulting in the indicator position not being accurately calculated. Similarly, the buttons might be rendered, but the font could take time to load, causing the indicator position to be calculated with the fallback font instead of the final one. Additionally, if the language is changed at runtime, it might result in a change in the buttons' length.

Long story short, observing the size of the container and repositioning the indicator when the size changes ensures that the indicator is always correctly positioned.

# Changes

- Use a `ResizeObserver` to set the indicator position.
- Mock `ResizeObserver` globally for vitest

# Screenshots

Before on main:

Capture d’écran 2025-01-21 à 07 19 52

After:

Capture d’écran 2025-01-21 à 07 19 37

# Tests

Tests of #563 should pass.

### DIFF
--- a/src/lib/components/Segment.svelte
+++ b/src/lib/components/Segment.svelte
@@ -78,18 +78,18 @@
 
   onDestroy(disconnectResizeObserver);
 
-  const resize = () => {
+  const observeSegmentResize = () => {
+    disconnectResizeObserver();
+
     if (isNullish(segment)) {
       return;
     }
-
-    disconnectResizeObserver();
 
     resizeObserver = new ResizeObserver(initIndicator);
     resizeObserver.observe(segment);
   };
 
-  $: segment, resize();
+  $: segment, observeSegmentResize();
 </script>
 
 <svelte:window on:resize={initIndicator} />
--- a/src/lib/components/Segment.svelte
+++ b/src/lib/components/Segment.svelte
@@ -2,7 +2,7 @@
   import { writable } from "svelte/store";
   import type { SegmentContext, SelectedSegment } from "$lib/types/segment";
   import { SEGMENT_CONTEXT_KEY } from "$lib/types/segment";
-  import { setContext, tick } from "svelte";
+  import { onDestroy, setContext, tick } from "svelte";
   import { isNullish, nonNullish } from "@dfinity/utils";
 
   export let selectedSegmentId: symbol | undefined = undefined;
@@ -59,7 +59,7 @@
     };
   };
 
-  $: selectedElement, (() => initIndicator())();
+  $: selectedElement, initIndicator();
 
   // TODO: support adding segmebt buttons dynamically
   let segmentsCount = 0;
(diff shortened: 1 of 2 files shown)


src/lib/components/Segment.svelte
```diff
@@ -59,14 +59,37 @@
...
+  onDestroy(disconnectResizeObserver);
+
+  const resize = () => {
+    if (isNullish(segment)) {
+      return;
+    }
+
+    disconnectResizeObserver();
+
+    resizeObserver = new ResizeObserver(initIndicator);
```
dskloetd, line 88: Would be nice if we could use anchor positioning instead but it looks like it's not widely supported yet.

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning/Using

src/lib/components/Segment.svelte
```diff
@@ -59,14 +59,37 @@
...
+  // Likewise, on mount, when the segment is bound and the indicator is set for the first time, the buttons might not be fully rendered in terms of size yet.
+  // Furthermore, if the content of the button dynamically changes—such as in applications that support runtime translations—their size might change.
+  // That is why, if the overall size of the component changes, we re-evaluate the position of the indicator.
+  let resizeObserver: ResizeObserver | undefined;
+
+  const disconnectResizeObserver = () => resizeObserver?.disconnect();
+
+  onDestroy(disconnectResizeObserver);
+
+  const resize = () => {
```
dskloetd, line 81: This name is confusing. It seems like this function would be used to resize the element. So I almost asked why we are recreating the observer each time the element is resized.
But it looks like what this really does is set up the resize observer, which can then we used for multiple resizes, correct?

src/lib/components/Segment.svelte
peterpeterparker: Oh nice! Yeah, it would probably require more resources since we would likely have to rewrite a few other things, hence the small improvement should do in my opinion.

src/lib/components/Segment.svelte
peterpeterparker: > This name is confusing.

Agree, renamed to `observeSegmentResize`. Not against better naming, let me know if you have better idea.

> are recreating the observer each time the element is resized

Because the function `unobserve` of the `ResizeObserver` requires the element to be unobserved, and with the reference system we are using, we do not track references—only pointing to the current reference. If the segment, for some reason, is detached and reattached to the DOM, we do not have a reference to the previous element; therefore, we cannot unobserve it. 

That's why we fully disconnect the observer, which unobserves all its observed elements.

That said, following your feedback, I reviewed the code again and moved the disconnection before checking if the segment is `nullish`. This way, if the segment was previously set and then became detached, we would remove the observer as well.

dskloetd (approved):

PR #567, opened 2025-01-27
chore: Playwright fix "Old Headless mode has been removed from the Chrome binary"

peterpeterparker:
# Motivation

The E2E tests starts failing in the CI (see for example this [job](https://github.com/dfinity/gix-components/actions/runs/12982647342/job/36202745642?pr=548)).

> pid=2146][err] Old Headless mode has been removed from the Chrome binary. Please use the new Headless mode (https://developer.chrome.com/docs/chromium/new-headless) or the chrome-headless-shell which is a standalone implementation of the old Headless mode (https://developer.chrome.com/blog/chrome-headless-shell).

The reason and the solution is explained in this Playwright issue: https://github.com/microsoft/playwright/issues/33566

Playwright documentation about Chromium: https://playwright.dev/docs/browsers#chromium

Issue I opened because I cannot run the Playwright command with the options: https://github.com/microsoft/playwright/issues/34487

# Changes

- Use channel `chromium` to download Chrome in the CI


playwright.config.ts
```diff
@@ -25,7 +25,7 @@ const config: PlaywrightTestConfig = {
   projects: [
     {
       name: "Google Chrome",
-      use: { ...devices["Desktop Chrome"], channel: "chrome" },
+      use: { ...devices["Desktop Chrome"], channel: "chromium" },
```
AntonioVentilii, line 28: I think "chronium" is the default, no?

AntonioVentilii (approved):
LGTM tks

playwright.config.ts
peterpeterparker: The linked resources on Playwright mention to explicitely change "chrome" to "chromium".

dskloetd:
In NNS dapp, I fixed it by updating the playwright dependency: https://github.com/dfinity/nns-dapp/pull/6207

peterpeterparker:
Thanks. PR replaced by #568 which bumps Playwright

PR #570, opened 2025-01-29
feat: adds autofocus to Input field

BonomoAlessandro:
# Motivation

We want to be able to set `focus` on the `Input` component. Since enabling `autofocus` is not recommended, we aim to make it possible to bind the input element. This way, we can set the focus programmaticcly and improve the user experience of our application.

# Changes

- enables to set focus on `Input` component


peterpeterparker (commented):
Minor comments.

Could you also list the property in the related table of the documentation?

https://github.com/dfinity/gix-components/blob/main/src/routes/(split)/components/input/%2Bpage.md

peterpeterparker (approved):
LGTM, thx

PR #571, opened 2025-01-29
chore: remove redundant statement for autocomplete in Input

AntonioVentilii:
# Motivation

There are two reactive statements that define `step` and `autocomplete`.

For step, they are identical, so we can remove them.

For `autocomplete`, one of the two ignores the prop `autocomplete` for currency input type. I think that the objective is to have this ONLY for type `number`.


peterpeterparker (approved):
Good catch! LGTM, thanks

PR #572, opened 2025-01-30
Reduce padding on mobile

mstrasinskis:
# Motivation

Currently, there are large side gaps on mobile that take up too much space. This PR improves the UI by reducing those gaps, allowing more content to be displayed on mobile.

Demo: https://qsgjb-riaaa-aaaaa-aaaga-cai.mstr-ingress.devenv.dfinity.network/

# Changes

- Remove extra padding and rounding from the `layout.content` mixin (it's used only in the `Content` and `SplitContent` components).
- Add top-left corner rounding to the `Content` and `SplitContent` components when the menu is open to maintain a consistent style.
- Decrease `Nav` padding to match the updated `Content` horizontal gaps.
- Use the default scrollbar style for mobile to prevent extra space allocation (a thin, auto-hiding bar that doesn't take up additional space).

# Screenshots

- Tested manually via the related `nns-dapp` pr - https://github.com/dfinity/nns-dapp/pull/6327

| Before | After |
|--------|--------|
| image | image |
| image | image |
| image | image |
| image | image |


src/lib/styles/global/layout.scss
```diff
@@ -10,10 +10,10 @@ main {
   max-width: var(--max-main-content-width);
   min-height: 100%;
 
-  padding: var(--padding-2x);
+  padding: var(--padding-2x) var(--padding);
 
-  @include media.min-width(large) {
-    padding: var(--padding-4x);
```
mstrasinskis, line 16: This was a redundant rule since it is overwritten on line 20.

e2e/progress-steps.e2e.ts-snapshots/Second-step-is-in-progress-1-Google-Chrome-linux.png
dskloetd, line 1: What changed in this image? I don't see anything and since it's not mobile I wouldn't have expected any changes here.

src/lib/components/Content.svelte
```diff
@@ -51,6 +51,7 @@
       box-shadow: var(--menu-selection-content-box-shadow);
       // Remove extra space because of menu selection touches the edge (when open)
       margin-left: 0;
+      border-top-left-radius: var(--border-radius-2x);
```
dskloetd, line 54: Is this to replace the border radius that was removed from `src/lib/styles/mixins/_layout.scss`?

Why is it only `-top-left` here?

src/lib/styles/global/scrollbar.scss
```diff
@@ -1,16 +1,25 @@
...
-  background: var(--scrollbar-thumb-color);
-  border: calc(var(--padding) * 2 / 3) var(--scrollbar-thumb-background) solid;
-  border-radius: 0.5em;
-  -webkit-border-radius: 0.5em;
-  background-clip: padding-box;
-}
+// Rely on the default scrollbar styles for mobile devices
+@include media.min-width(medium) {
+  html {
+    scrollbar-width: thin; /* Firefox */
```
dskloetd, line 6: Why does it says "Firefox"?
According to [this](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width), `scrollbar-width` is available on all major browsers.

src/lib/components/SplitContent.svelte
```diff
@@ -69,6 +69,7 @@
       box-shadow: var(--menu-selection-content-box-shadow);
       // remove extra space because of menu selection touches the edge (when open)
       margin-left: 0;
+      border-top-left-radius: var(--border-radius-2x);
```
dskloetd, line 72: Why only `-top-left`?

src/lib/components/Header.svelte
```diff
@@ -26,8 +26,10 @@
   @use "../styles/mixins/media";
 
   header {
-    --toolbar-padding: 0 var(--padding-2x);
-
+    --toolbar-padding: 0;
+    @include media.min-width(medium) {
+      --toolbar-padding: 0 var(--padding-2x);
```
dskloetd, line 31: Do I read correctly that `--toolbar-padding` is zero on small and large but not medium? Why?

e2e/progress-steps.e2e.ts-snapshots/Second-step-is-in-progress-1-Google-Chrome-linux.png
mstrasinskis: Looks like some unrelated anti-aliasing or compression artifacts:
Highlighted Different Pixels (Red)
I ran the action twice with the same result.

src/lib/components/Content.svelte
mstrasinskis: Since we no longer display the bottom border radius, this is the only visible area

Before
image


After
image

src/lib/components/Header.svelte
mstrasinskis: Because on large screens only the left button is shown (aligned to the right), and the title is hidden since it's in the nav menu.

image

image

image

src/lib/styles/global/scrollbar.scss
mstrasinskis: Good catch! I completely overlooked that and forgot to remove it after testing.

e2e/progress-steps.e2e.ts-snapshots/All-steps-are-completed-1-Google-Chrome-linux.png
dskloetd, line 1: I thought we allow small changes in screenshot tests and the changes here are literally not visible.
Does the test actually fail if you revert the screenshot change?

dskloetd (approved):

e2e/progress-steps.e2e.ts-snapshots/All-steps-are-completed-1-Google-Chrome-linux.png
mstrasinskis: After resetting the screenshots, only one image is different (which is expected). I now suppose that this was a result of updating the snapshots multiple times during development.

**What fails after reseting the snapshots:**
https://github.com/dfinity/gix-components/actions/runs/13130907651/job/36635693219?pr=572

```
...
- 20481 pixels (ratio 0.08 of all image pixels) are different.
...
  1 failed
    [Google Chrome] › e2e/tooltip.e2e.ts:5:1 › Should render tooltip ───────────────────────────────
  45 passed (1.0m)
Error: Process completed with exit code 1.
```

dskloetd: Thanks for checking! I agree with the hypothesis.

PR #574, opened 2025-02-04
feat: use directive transition wrappers

peterpeterparker:
# Motivation

Implements the transition directives wrappers provided in #573 for the transition that lead the [CI](https://github.com/dfinity/nns-dapp/actions/runs/13129931873/job/36632946426) to fail in the NNS dapp Svelte v5 branch.

# Changes

- Instead of using `svelte/transition`, use the wrappers that skips the transitions for testing purposes

# Tests

I asserted this changes has no impact on the test suite of NNS dapp by integrating a manual package and running the CI successfully in https://github.com/dfinity/nns-dapp/pull/6336.

We will have to mock the `ResizeObserver` though but, that's something we need for Svelte v5 testing anyway https://github.com/dfinity/nns-dapp/pull/6337

### DIFF
--- a/src/lib/components/Backdrop.svelte
+++ b/src/lib/components/Backdrop.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { fade } from "svelte/transition";
+  import { testSafeFade } from "$lib/directives/transition.directives";
   import { createEventDispatcher } from "svelte";
   import { i18n } from "$lib/stores/i18n";
   import { handleKeyPress } from "$lib/utils/keyboard.utils";
@@ -18,8 +18,8 @@
   role="button"
   tabindex="-1"
   aria-label={$i18n.core.close}
-  in:fade|global={{ duration: FADE_IN_DURATION }}
-  out:fade|global={{ duration: FADE_OUT_DURATION }}
+  in:testSafeFade|global={{ duration: FADE_IN_DURATION }}
+  out:testSafeFade|global={{ duration: FADE_OUT_DURATION }}
   class="backdrop"
   class:visible={!invisible}
   on:click|stopPropagation={close}
--- a/src/lib/components/BusyScreen.svelte
+++ b/src/lib/components/BusyScreen.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { fade } from "svelte/transition";
+  import { testSafeFade } from "$lib/directives/transition.directives";
   import { busy, busyMessage } from "$lib/stores/busy.store";
   import Spinner from "$lib/components/Spinner.svelte";
   import { nonNullish } from "@dfinity/utils";
@@ -7,7 +7,7 @@
 
 <!-- Display spinner and lock UI if busyStore is not empty -->
 {#if $busy}
-  <div data-tid="busy" transition:fade|global>
+  <div data-tid="busy" transition:testSafeFade|global>
     <div class="content">
       {#if nonNullish($busyMessage)}
         <p>{$busyMessage}</p>
--- a/src/lib/components/Modal.svelte
+++ b/src/lib/components/Modal.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { fade } from "svelte/transition";
+  import { testSafeFade } from "$lib/directives/transition.directives";
   import { createEventDispatcher } from "svelte";
   import { i18n } from "$lib/stores/i18n";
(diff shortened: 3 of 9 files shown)


src/lib/components/Backdrop.svelte
dskloetd, line 1: Just wondering, have you considered mocking the `"svelte/transition"` package to return the test safe transitions so we don't need to change the production code for the test purpose?

peterpeterparker: I also tried mocking the entire `svelte/transition`, and while it seemed to work better at first, further testing rounds revealed that it did not resolve all issues. Errors were still being thrown due to the lack of an `animate` mock. So, I tried combining both approaches, but it did not worked out neither. Since the author of Svelte Testing discourages mocking `svelte/transition` (this [comment](https://github.com/testing-library/svelte-testing-library/issues/416#issuecomment-2573404786)), I ultimately implemented this as it felt like the only viable solution.

peterpeterparker: That said, reading your comment again and understanding what you mean, I did not try to mock `svelte/transition` with our  new function s`safeTransition...` . I can give it a try.

dskloetd: Would be great if you could try. Approving in case it doesn't work.

dskloetd (approved):

src/lib/components/Backdrop.svelte
peterpeterparker: I just tried following in NNS dapp `vitest.setup`:

```
vi.mock("svelte/transition", async (importOriginal) => {
  const svelteStoreModule =
      await importOriginal<typeof import("svelte/transition")>();

  return {
    ...svelteStoreModule,
    fade: testSafeFade,
    fly: testSafeFly,
    scale: testSafeScale
  };
});
```

But when I run the test I get following error:

```

 FAIL  src/tests/lib/pages/NnsWallet.spec.ts [ src/tests/lib/pages/NnsWallet.spec.ts ]
Error: [vitest] There was an error when mocking a module. If you are using "vi.mock" factory, make sure there are no top level variables inside, since this call is hoisted to top of the file. Read more: https://vitest.dev/api/vi.html#vi-mock
 ❯ node_modules/@dfinity/gix-components/dist/components/Backdrop.svelte:2:249
      6| 
      7|   export let disablePointerEvents = false;
      8|   export let invisible = false;
       |          ^
      9| 
     10|   const dispatch = createEventDispatcher();

Caused by: ReferenceError: Cannot access '__vi_import_12__' before initialization
 ❯ vitest.setup.ts:110:11
 ❯ node_modules/@dfinity/gix-components/dist/components/Backdrop.svelte:2:249
```

console.log of `svelteStoreModule` prints:

```
stdout | src/tests/lib/pages/NnsWallet.spec.ts
[Object: null prototype] [Module] {
  blur: [Getter],
  fade: [Getter],
  fly: [Getter],
  slide: [Getter],
  scale: [Getter],
  draw: [Getter],
  crossfade: [Getter]
}
```

Do I missed something?

dskloetd: You need to put the import for `testSafeFade` etc. on the first line of the file because of the hoisting.

peterpeterparker: Do you mean first line of the `vitest.config`?

```
1 import {testSafeFade, testSafeFly, testSafeScale} from "@dfinity/gix-components";
2 other imports
3 etc

etc
```

Lead to the same issue. Probably a stupid question, sorry for that.

Note: if you want to give it a shot, the branch is there https://github.com/dfinity/nns-dapp/pull/6020. Up-to-date minus this PR and the config we are discussing.

dskloetd: `vitest.setup.ts`, not `config`, right?

If that doesn't work, I also don't know. Thanks for trying.
It's what I did here, for example: https://github.com/dfinity/nns-dapp/blob/2ca691f2291cac9df3ed7f522c77839d11907bae/frontend/src/tests/lib/modals/transaction/QrWizardModal.spec.ts#L1

peterpeterparker: Yes correct `vitest.setup.ts`, sorry for the typo.

I understood correctly and indeed it does not seem to work or, at least, leads to another issue.

Capture d’écran 2025-02-04 à 14 14 41

We can always revert this PR if we ever find out. Thanks for the review and suggestions!

PR #582, opened 2025-02-11
feat: refresh method in themeStore

AntonioVentilii:
# Motivation

We would like to have a `refresh` method in the `themeStore`. It is useful for example, if the theme is changed in the local storage and we need to "refresh" the store: for example, we have the case in OISY where we "set" the theme to `SYSTEM`, basically deleting the theme in local storage.

# Changes

- Extract an util `getCurrentTheme` to get the current theme set in the store: this code is already inside `initTheme`.
- Create sub-function `setTheme` inside `initialTheme` that is re-using the code of the `select` function.
- Reusing the sub-function above for the method `refresh` in combination with `getCurrentTheme`.

# Screenshots

Added a few tests for `themeStore` and `initThemeStore`.

### DIFF
--- a/src/lib/stores/theme.store.ts
+++ b/src/lib/stores/theme.store.ts
@@ -1,19 +1,24 @@
 import type { Theme } from "$lib/types/theme";
-import { applyTheme, initTheme } from "$lib/utils/theme.utils";
+import { applyTheme, getCurrentTheme, initTheme } from "$lib/utils/theme.utils";
+import { nonNullish } from "@dfinity/utils";
 import { writable } from "svelte/store";
 
-const initialTheme: Theme | undefined = initTheme();
-
 export const initThemeStore = () => {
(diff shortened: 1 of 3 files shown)


AntonioVentilii:
@peterpeterparker shall I break this in smaller PRs? If yes, any suggestion?

peterpeterparker (changes requested):
I don't understand your PR — it doesn't seem to implement the function we discussed. I might be missing something, but is it designed this way because some logic remains in OISY? Is that the idea?

To be clear, this is the feature logic I would have expected:  https://github.com/dfinity/gix-components/pull/585

AntonioVentilii:
@peterpeterparker Ah! I completely misunderstood then! Yes, the idea was to leave some logic there in OISY, but if that is not the case, for sure let's go with your PR

AntonioVentilii:
As discussed offline, I will close this and continue with the correct implementation in :

- #585

PR #584, opened 2025-02-12
chore: interfaces for theme store

peterpeterparker:
# Motivation

Improve the types of the theme store by declaring interfaces as we commonly do nowadays.

### DIFF
--- a/src/lib/stores/theme.store.ts
+++ b/src/lib/stores/theme.store.ts
@@ -1,11 +1,17 @@
 import type { Theme } from "$lib/types/theme";
 import { applyTheme, initTheme } from "$lib/utils/theme.utils";
-import { writable } from "svelte/store";
+import { writable, type Readable } from "svelte/store";
+
+export type ThemeStoreData = Theme | undefined;
+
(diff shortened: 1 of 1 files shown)


src/lib/stores/theme.store.ts
```diff
@@ -1,11 +1,17 @@
...
-import { writable } from "svelte/store";
+import { writable, type Readable } from "svelte/store";
+
+export type ThemeStoreData = Theme | undefined;
+
+export interface ThemeStore extends Readable<ThemeStoreData> {
+  select: (theme: Theme) => void;
+}
 
 const initialTheme: Theme | undefined = initTheme();
```
yhabib, line 11: [nitpick]
```suggestion
const initialTheme: ThemeStoreData = initTheme();
```

yhabib (approved):
Thanks

src/lib/stores/theme.store.ts
peterpeterparker: I developed the interface in another PR and missed copying this line over. Good catch, thanks!

PR #586, opened 2025-02-12
feat: revert "use directive transition wrappers"

peterpeterparker:
# Motivation

Revert #574 because thanks to a different way of mocking the transitions for test purposes (without vite), those workaround are not needed anymore.

# Resources

- Mock tested in https://github.com/dfinity/nns-dapp/pull/6404
- New solution to mock the animation for test purposes with Svelte v5 providing by Testing Library in https://github.com/testing-library/svelte-testing-library/issues/284#issuecomment-2651620976

# Changes

- Revert #574 and use Svelte transitions


dskloetd (approved):

peterpeterparker:
Not sure exactly why but, this PR leads to a pixel change in the test `Menu can be collapsed`.

image

Visually seems acceptable to.

PR #589, opened 2025-02-14
Hide top on scroll

mstrasinskis:
# Motivation

To improve the mobile experience by maximizing screen space, this update hides the header when the user scrolls down.

Initially, the approach was to hide the header on scroll down and show it on scroll up. However, this caused glitches due to container resizing because of SplitPage/padding: 0. Instead, a more reliable solution was chosen: hiding the header when the scrollable content starts leaving the visible viewport.

demo: https://qsgjb-riaaa-aaaaa-aaaga-cai.mstr-ingress.devenv.dfinity.network/

# Changes

- Introduced a new ScrollSentinel component that observes when it leaves the viewport and updates layoutContentTopHidden.
- The layoutContentTopHidden value is now used to toggle the header’s visibility (to removed SplitPage top padding and the header when the value is true)

- Tested on iOS and Android.

# Screenshots

https://github.com/user-attachments/assets/72df4b50-f0b6-4a80-862c-4b6f48fe95fd

### The glitch of the first approach

https://github.com/user-attachments/assets/e3b3c78e-7185-42a9-a92b-a4b40e127b35


src/lib/components/Content.svelte
```diff
@@ -25,7 +31,11 @@
     <slot name="toolbar-end" slot="toolbar-end" />
   </Header>
 
-  <div class="scrollable-content" class:open={$layoutMenuOpen}>
+  <div
+    class="scrollable-content"
+    class:open={$layoutMenuOpen}
+    on:scroll={handleScroll}
```
peterpeterparker, line 37: I know this PR is a draft, but as early feedback, why are you hooking into scroll? It's a bit suboptimal in terms of performance. Assuming you want to hide something at the top on scroll, can't you use an Intersection Observer?

mstrasinskis: Thanks for your early feedback. The initial plan was to hide on scroll down and display it back on scroll up. Not sure it's doable with the Intersection Observer.

peterpeterparker: Ah to hide when it should be still visible on screen? Indeed it's a bit more tricky. Nvm my comment then. Thought you were building something like this.

Capture d’écran 2025-02-17 à 17 40 02
Capture d’écran 2025-02-17 à 17 39 47

mstrasinskis: Actually something like this https://juno.build/docs/guides/nextjs

peterpeterparker: Gotcha, that's Docusaurus.

But maybe like this too https://recyclingcredit.org/ ?

That's dapp I built with Svelte for the foundation - i.e. we have the source code, if interesting.

src/lib/components/Header.svelte
```diff
@@ -2,11 +2,12 @@
   import Toolbar from "$lib/components/Toolbar.svelte";
   import MenuButton from "$lib/components/MenuButton.svelte";
   import Back from "$lib/components/Back.svelte";
+  import { layoutContentTopHidden } from "$lib/stores/layout.store";
 
   export let back = false;
 </script>
 
-<header data-tid="header-component">
+<header data-tid="header-component" class:is-hidden={$layoutContentTopHidden}>
```
peterpeterparker, line 10: opiniated probably: I would just call it `hidden`, it's a design style class, not a logical element.

src/lib/components/Header.svelte
```diff
@@ -33,5 +34,16 @@
     @include media.min-width(large) {
       --toolbar-padding: 0;
     }
+
+    &.is-hidden {
+      opacity: 0;
+      transform: translateY(-100%);
```
peterpeterparker, line 40: Works on iOS? #neverTrustSafari

src/lib/components/ScrollSentinel.svelte
```diff
@@ -0,0 +1,22 @@
...
+  // To observe when the top leaves the view
+  let element: HTMLElement;
+
+  onMount(() => {
+    const observer = new IntersectionObserver(
+      ([entry]) => layoutContentTopHidden.set(!entry.isIntersecting),
+      { root: scrollContainer, threshold: 0 },
+    );
+    observer.observe(element);
+    return () => observer.disconnect();
```
peterpeterparker, line 18: dumb question: same effect as registering an on destroy?

src/lib/components/ScrollSentinel.svelte
```diff
@@ -0,0 +1,22 @@
...
+  });
+</script>
+
+<div data-tid="sentinel" bind:this={element}></div>
```
peterpeterparker, line 22: What do you think about using the same pseudo hidden style as the placeholder of the InfiniteScroll cmp here (if applicable)?

```
.intersection-observer-target {
    width: 0;
    height: 0;
    opacity: 0;
    visibility: hidden;
  }
```

src/lib/components/SplitPane.svelte
```diff
@@ -18,7 +21,7 @@
 
 <svelte:window bind:innerWidth />
 
-<div class="split-pane">
+<div class="split-pane" class:is-header-hidden={$layoutContentTopHidden}>
```
peterpeterparker, line 24: ditto about `is-`

src/lib/components/ScrollSentinel.svelte
mstrasinskis: I expect the same effect as [onUnmount](https://svelte.dev/docs/svelte/lifecycle-hooks#onMount).

src/lib/components/Header.svelte
mstrasinskis: Looks good 👍

src/lib/components/ScrollSentinel.svelte
mstrasinskis: Thank you! That makes sense to avoid accidentally making it visible.

src/lib/components/ScrollSentinel.svelte
peterpeterparker: Nice! 

To be sure I double checked the doc of Svelte v4: https://v4.svelte.dev/docs/svelte#onmount

src/lib/components/ScrollSentinel.svelte
peterpeterparker: Cool. Maybe after this PR we should extract a mixin for it?

src/lib/components/SplitPane.svelte
```diff
@@ -42,6 +45,18 @@
...
+    transition: padding-top var(--animation-time-normal) ease;
+
+    &.header-hidden {
+      padding-top: 0;
+      // Reset on tablet+
+      @include media.min-width(medium) {
+        padding-top: var(--split-pane-content-top-offset);
+      }
+      @include media.min-width(large) {
+        padding-top: var(--header-offset, 0px);
```
peterpeterparker, line 57: `0` instead of `0px`? Just asking because I know sometimes the linter complains about it.

peterpeterparker (approved):
LGTM, thx

PR #590, opened 2025-02-18
chore: use eslint-config-oisy-wallet / eslint v9

peterpeterparker:
# Motivation

Instead of just migrating to ESLint v9, I think we could migrate to using [eslint-config-oisy-wallet](https://github.com/dfinity/eslint-config-oisy-wallet), as we do in OISY and other repositories. It's a somewhat opinionated library, but to some extent, it doesn't enforce many more rules than the ones we are already using. Given that the delta isn't too big, and if the team agrees, I think it would be handy to reuse the same strategies for the projects we collaborate on.

# Changes

# Screenshots


peterpeterparker:
Discussed https://github.com/dfinity/gix-components/pull/631#discussion_r2092292532

PR #598, opened 2025-02-28
Chip group

mstrasinskis:
# Motivation

For the new table sorting UI in the NNS Dapp, we need a ChipGroup component.

# Changes

- New Chip and ChipGroup components.

# Screenshots

Many CI screenshots have been updated due to the new component in the side menu. Additionally, one mobile screenshot was updated due to the absence of the top header, which was introduced some time ago, though I’m not sure why it happened now.

**Light theme**
image

**Dark theme**
image


src/lib/components/Chip.svelte
```diff
@@ -0,0 +1,53 @@
...
+  export let selected: boolean;
+
+  const dispatch = createEventDispatcher();
+</script>
+
+<button
+  class="chip"
+  data-tid="chip-component"
+  class:selected
+  on:click={() => dispatch("nnsClick", id)}>{label}</button
```
yhabib, line 15: wdyt of?
```
  aria-pressed={selected}
  role="option"
```

src/lib/components/ChipGroup.svelte
```diff
@@ -0,0 +1,38 @@
...
+  const onChipClick = ({ detail: selectedId }: CustomEvent<string>) => {
+    chips = chips.map((chip) => ({
+      ...chip,
+      selected: chip.id === selectedId,
+    }));
+    dispatch("nnsSelect", selectedId);
+  };
+</script>
+
+<div data-tid={testId} class="chip-group">
```
yhabib, line 26: wdyt of
```
  role="radiogroup" 
  aria-label="Options"
```

yhabib (approved):
LGTM left a few minor comments. 
I hope we aren't planning to use it with so many items 😅. If we expect to have more than 6 or 7, we should consider using a dropdown instead 🙈.

src/lib/components/Chip.svelte
mstrasinskis: Good point, updated!

PR #599, opened 2025-03-03
test: use events instead of $on for theme listener

peterpeterparker:
# Motivation

Using `$on` is incompatible with Svelte v5.

# Changes

- Update theme listener test to use the existing `events` property of the render utils.


mstrasinskis (approved):
Thank you!

PR #602, opened 2025-03-12
Fix scss warning

mstrasinskis:
# Motivation

Currently, when the _button SCSS mixin is applied, the disabled states are always added. Because of this, in cases where they are not needed (e.g., in the Chip component), the linter produces warnings:
```
Warn: Unused CSS selector ".chip[disabled]" (svelte)

<style lang="scss">
  @use "../styles/mixins/effect";
```

# Changes

- Add flag to the _button mixin to skip disable styles.
- Use the flag in the Chip to avoid warnings.

# Screenshots

No visual or logical changes.

### DIFF
--- a/src/lib/components/Chip.svelte
+++ b/src/lib/components/Chip.svelte
@@ -23,7 +23,7 @@
   @use "../styles/mixins/button";
 
   .chip {
-    @include button.base;
+    @include button.base(false);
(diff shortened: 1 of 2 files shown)


AntonioVentilii:
Came looking for this! Thank you!

PR #603, opened 2025-03-14
feat: Collapsible content new prop to avoid overflow hidden

daviddecentage:
# Motivation

Due to some issues with borders and outlines being cut for collapsible content, I added the possibility to apply an offset to be able to avoid content being cut off.

# Changes

- Added `offsetContent` prop to collapsible component
- Added styling for two new classes:
- - `.offset-content` for normal collapsibles
- - `.offset-content-scroll` for when maxContentHeight is specified as well, because the margins behave a bit differently when content is scrollable (which can only happen if maxContentHeight is set)
- Updated docs

# Screenshots

In the screenshot we can see that the closed state is not affected, and the black outline is visible by setting `offsetContent` property.

image

image


AntonioVentilii:

Thanks for the contribution!

PR #605, opened 2025-03-18
feat(ui): add bottom slot to input component

yhabib:
# Motivation

We want to display the USD value in transaction forms within the nns-dapp. To achieve this, we need to extend the input to include a "bottom" section with custom styles.

Example from the nns-dapp:

Screenshot 2025-03-18 at 17 51 34

# Changes

- Expose a new slot in the `Input` field

# Screenshots

Screenshot 2025-03-18 at 17 47 42

Screenshot 2025-03-18 at 17 48 02

### DIFF
--- a/src/lib/components/Input.svelte
+++ b/src/lib/components/Input.svelte
@@ -171,34 +171,36 @@
       <slot name="end" />
     </div>
   {/if}
-  <div class="input-field" class:withBottom={displayBottom}>
-    <input
-      bind:this={inputElement}
-      data-tid={testId}
-      type={currency ? "text" : inputType}
-      {required}
-      {spellcheck}
-      {name}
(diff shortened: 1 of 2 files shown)


src/lib/components/Input.svelte
```diff
@@ -168,32 +171,40 @@
...
+      />
+      {#if displayInnerEnd}
+        <div class="inner-end-slot">
+          <slot name="inner-end" />
+        </div>
+      {/if}
+    </div>
+
+    {#if displayBottom}
+      <div class="bottom-slot">
```
mstrasinskis, line 206: Why is this wrapper/class needed?

mstrasinskis (approved):
Thanks!

PR #613, opened 2025-04-01
chore(ci): upgrade github actions runner to ubuntu-22.04

yhabib:
# Motivation

Gix uses an Ubuntu runner that will be deprecated on April 1st. A Slack [message](https://dfinity.slack.com/archives/CH4CADCJX/p1740558752499169) pointed to some repositories, but gix-components was overlooked.

We have some issues in the pipeline with open PRs. For example: https://github.com/dfinity/gix-components/actions/runs/14200871834

# Changes

- Upgraded the version to `ubuntu-22.04`.

# Tests

- CI should pass

# Todos

- [ ] Add entry to changelog (if necessary).
Not necessary.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -8,11 +8,11 @@ on:
 
 jobs:
   formatting:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     # In order to trigger other workflows after committing formatting changes, we need
     # to use the PR Automation App. This secret is not available for external
-    # contributors. So on PRs that can't acces the secret, we don't commit changes and instead just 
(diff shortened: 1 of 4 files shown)


peterpeterparker (commented):
Should we use `ubuntu-latest`?

yhabib:
I was trying to minimize the jump. We couldn't bump n a-dapp to the latest, but it was probably related to the node version.
I'm going to revisit it in a couple of months for the n a-dapp, and I thought about tackling it for Giza then as well.

peterpeterparker:
Sounds good!

Just noticed ic-js uses `ubuntu-20.04`, can you provide a PR for that other repo as well?

peterpeterparker (approved):
LGTM, thx

PR #616, opened 2025-04-05
feat(icons): add size prop for launchpad and token icons

yhabib:
# Motivation

The nns-dapp wants to use the `IconLaunchpad` and `IconTokens` in different sizes. This PR introduces a `size` prop, similar to the approach used for other icons. To maintain backward compatibility, it does not use `DEFAULT_ICON_SIZE`; instead, it employs a local variable to set the size to 40 px.

# Changes

- Expose the new `size` prop in `IconLaunchpad` and `IconTokens`

# Screenshots

Not necessary.


mstrasinskis (approved):
Thanks!

PR #620, opened 2025-04-16
feat(modal): Adds slots for the left and right side of modal title

FlavioInacta:
# Motivation

To add custom buttons to the modal. This will be used mainly for mobile view, but also to add a button on top left to add a go back button.

# Changes

Adds two slots `header-left` and `header-right`.


daviddecentage (approved):
Thats gonna be very useful! LGTM, thanks.

AntonioVentilii (changes requested):
Did you run the E2E tests for this one?

src/lib/components/Modal.svelte
```diff
@@ -222,20 +231,29 @@
 
     position: relative;
 
+    .header-left {
+      display: flex;
+      justify-content: left;
```
peterpeterparker, line 236: Any particular reason to use `left` and `right`? We generally use `flex-start` and `flex-end`, it was also the case of the style you remove, would those work out here?

src/lib/components/Modal.svelte
```diff
@@ -222,20 +231,29 @@
...
+      align-items: center;
+    }
+
     button {
       display: flex;
       justify-content: center;
       align-items: center;
       padding: 0;
-
-      justify-self: flex-end;
```
peterpeterparker, line 237: I wonder if this will introduce a change in behavior. We might not catch it with the test suite of Gix-cmp. So if we proceed with this PR, I think it would be best to build the next version and immediately provide a PR to the NNS dapp to ensure it remains backwards compatible. Could you do that?

FlavioInacta: What do you mean with "behaviour"? It should render exactly the same as before.
I have no idea how to release nor what the nns team is, but I will find someone who can help me.

src/lib/components/Modal.svelte
FlavioInacta: Oh, thats wrong. Thank you. Should be `flex-end` and `flex-start`.

src/lib/components/Modal.svelte
peterpeterparker: For example the click area will be a bit different

Capture d’écran 2025-04-17 à 12 17 50
Capture d’écran 2025-04-17 à 12 17 40

peterpeterparker: But as I said, if we can integrate it right after merge and validate the test suite of  NNS dapp then I guess we will be fine.

src/lib/components/Modal.svelte
```diff
@@ -62,16 +62,25 @@
     >
       {#if showHeader}
         <div class="header">
+          <div class="header-left">
```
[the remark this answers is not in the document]
FlavioInacta: Done

src/lib/components/Modal.svelte
FlavioInacta: Fixed

peterpeterparker (commented):
Looks good. The CI reports an issue, you probably have to run `npm run format` or comply with `npm run lint` (did not check the logs). Can you have a look?

peterpeterparker:
The E2E tests are actually failing in the new PR https://github.com/dfinity/gix-components/actions/runs/14515913608/job/40724821614?pr=621. Flaky or related to the changes?

Btw. should we close this PR to continue with the new one?

FlavioInacta:
@peterpeterparker Seems like the new one is not needed. After I have access to the repos the PR from my fork seems fine. (Yes looks flaky to me, I did not change anything there)

AntonioVentilii (approved):
Unblocking

PR #622, opened 2025-04-22
feat(input): Adds autofocus property to input component

FlavioInacta:
# Motivation

In OISY, the autofocus is used several times with a custom implementation. To be able to use the browsers implementation of autofocus, this PR exposes the `autofocus` property.

Compare: https://github.com/dfinity/oisy-wallet/pull/5901

# Changes

- Exposes the `autofocus` property

### DIFF
--- a/src/lib/components/Input.svelte
+++ b/src/lib/components/Input.svelte
@@ -16,6 +16,7 @@
   export let decimals = 8;
   export let ignore1Password = true;
   export let inputElement: HTMLInputElement | undefined = undefined;
+  export let autofocus = false;
 
   const dispatch = createEventDispatcher();
 
@@ -173,6 +174,7 @@
   {/if}
   <div class:with-bottom={displayBottom}>
     <div class="input-field">
+      <!-- svelte-ignore a11y_autofocus -->
       <input
         bind:this={inputElement}
         data-tid={testId}
@@ -188,6 +190,7 @@
         {placeholder}
         {max}
         {autocomplete}
+        {autofocus}
         on:blur
         on:focus
         on:input={handleInput}
--- a/src/tests/lib/components/Input.svelte.spec.ts
+++ b/src/tests/lib/components/Input.svelte.spec.ts
@@ -1,7 +1,7 @@
 import Input from "$lib/components/Input.svelte";
 import { assertNonNullish, isNullish, nonNullish } from "@dfinity/utils";
-import { fireEvent, render } from "@testing-library/svelte";
 import { tick } from "svelte";
+import { fireEvent, render, waitFor } from "@testing-library/svelte";
 import InputElementTest from "./InputElementTest.svelte";
 import InputTest from "./InputTest.svelte";
 import InputValueTest from "./InputValueTest.svelte";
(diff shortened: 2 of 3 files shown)


src/lib/components/Input.svelte
```diff
@@ -188,6 +190,7 @@
         {placeholder}
         {max}
         {autocomplete}
+        {autofocus}
```
yhabib, line 193: There was [this](https://github.com/dfinity/gix-components/pull/570) previous approach to this where it mentions that enabling `autofocus` is not recommended, did something change?

FlavioInacta: This does not seem to be an autofocus implementation. But yes, it is used to manually focus an element. idk. Seems wrong to me. @BonomoAlessandro can you explain? Where there discussions about using `autofocus` attribute?

peterpeterparker: I'm not entirely sure, but I think the note about "autofocus not recommended" was meant in the context of the specific feature that was being built — i.e. the feature involved setting autofocus on an input within a particular form. As far as I remember, that doesn't always work reliably, which is why an input reference was ultimately used to set focus programmatically.

In other cases, autofocus might still be useful, so I think it's fine to include it as an optional attribute.

src/tests/lib/components/Input.svelte.spec.ts
```diff
@@ -554,6 +554,36 @@ describe("Input", () => {
...
+  it("should autofocus the input when autofocus is true", async () => {
+    const { container } = render(Input, {
+      props: { ...props, autofocus: true },
+    });
+
+    const input: HTMLInputElement | null = container.querySelector("input");
+    expect(input).not.toBeNull();
+
+    // In testing environments, we need to wait for the DOM to update
+    await tick();
```
peterpeterparker, line 566: waitFor instead of tick?

src/tests/lib/components/Input.svelte.spec.ts
```diff
@@ -554,6 +554,36 @@ describe("Input", () => {
...
+
+    // Check if the input is the active element
+    expect(input === document.activeElement).toBe(true);
+  });
+
+  it("should not focus by default", async () => {
```
peterpeterparker, line 572: should we add an assertion to ensure the attribute is not rendered as well?

peterpeterparker (commented):
Looks good, minor questions about the tests.

Do you want to update the documentation in this PR or in another? As you rather like.

src/tests/lib/components/Input.svelte.spec.ts
FlavioInacta: Done

src/tests/lib/components/Input.svelte.spec.ts
```diff
@@ -467,9 +466,9 @@ describe("Input", () => {
       // svelte does not update the dom immediately
       expect(input.value).toBe("");
 
-      await tick();
-
-      expect(input.value).toBe("0.00000001");
+      await waitFor(() => {
```
peterpeterparker, line 469: maybe revert this one since it's not stricto sensu related to the goal of the PR

peterpeterparker (approved):
LGTM,  thx

Just added a comment about the test if you can have a look before moving forward. Already approving.

src/tests/lib/components/Input.svelte.spec.ts
FlavioInacta: Oh, right. Just replaced all ticks. ;)

PR #623, opened 2025-04-22
feat(input): Add clear button to input component

FlavioInacta:
# Motivation

In order to easily remove the content of a input field, an optional clear button can be displayed.

# Changes

Adds a clear button when `showClearButton` is set to `true` on the input field.

# Screenshots

Without any value:

Screenshot_2025-04-22_14-29-38

Button is displayed when the input field has some value:

Screenshot_2025-04-22_14-29-57

### DIFF
--- a/src/lib/components/Input.svelte
+++ b/src/lib/components/Input.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import { IconClose } from "$lib/icons";
   import { isNullish, nonNullish } from "@dfinity/utils";
   import { createEventDispatcher } from "svelte";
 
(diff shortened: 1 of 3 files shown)


src/lib/components/Input.svelte
```diff
@@ -198,6 +204,11 @@
       {#if displayInnerEnd}
         <div class="inner-end-slot">
           <slot name="inner-end" />
+          {#if showClearButton && value}
```
yhabib, line 207: Do we really need this new prop? Why not utilize the `inner-end` slot directly from the consumer of this component (Oisy)? If you use it frequently, you could create your own version there. 
If we do need this new prop, should it work together with the `inner-end` slot, or should only one be used?

AntonioVentilii: @FlavioInacta maybe there was a misunderstanding: i didn't want this specific part here, only the `autofocus`, see my comment https://github.com/dfinity/oisy-wallet/pull/5901#issuecomment-2814650384

FlavioInacta:
Will just add it to oisy wallet

PR #625, opened 2025-04-28
fix(deps): upgrade sveltekit to resolve icons page serialization

yhabib:
# Motivation

The upgrade to Svelte 5 broke the [icons page](https://gix.design/icons) due to a regression issue caused by Svelte Kit deserializing the payload from the backend file.

Svelte Kit has addressed this issue in the [latest version](https://github.com/sveltejs/kit/releases/tag/%40sveltejs%2Fkit%402.20.7).

# Changes

- Upgrade SvelteKit to the latest version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "@junobuild/config": "^0.0.14",
         "@playwright/test": "^1.51.0",
         "@sveltejs/adapter-static": "^3.0.8",
-        "@sveltejs/kit": "^2.20.4",
+        "@sveltejs/kit": "^2.20.7",
         "@sveltejs/package": "^2.3.10",
         "@sveltejs/vite-plugin-svelte": "^5.0.3",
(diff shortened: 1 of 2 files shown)


peterpeterparker (approved):
LGTM, thx

Should we urgently upgrade SvelteKit in NNS dapp or OISY or is the issue limited to the showcase page of Gix-Cmp?

Side note: I guess we discussed it in the past but, never implemented it. Maybe we really needed an E2E screenshot test over the icons page

yhabib:
@peterpeterparker the nns-dapp was upgraded last week, but we did not encounter this issue. The [bug](https://github.com/sveltejs/kit/pull/13709) seems to be related to serializing the server data. This was the only page from Gix that was affected. The icons themselves were fine; only the demo page had problems.

peterpeterparker:
Good to hear, thx

PR #629, opened 2025-05-15
feat!: migrating `InfiniteScroll` to Svelte 5

AntonioVentilii:
# Motivation

We start the Svelte 5 migration with component InfiniteScroll, that is more prone to cause incompatibility issues with projects that already use Svelte 5 components.

Note that I am explicitly set the prop `children` as non-optional, since I imagine that this component shall be used only with children.

BREAKING CHANGE: the interface changes.

Credits to @peterpeterparker from [Juno's component `InfiniteScroll`](https://github.com/junobuild/juno/blob/cd44adace1ff530cc8c2f544bfbc77b25aace88a/src/frontend/src/lib/components/ui/InfiniteScroll.svelte#L5)

### DIFF
--- a/src/lib/components/InfiniteScroll.svelte
+++ b/src/lib/components/InfiniteScroll.svelte
@@ -3,7 +3,7 @@
   import { isNullish } from "@dfinity/utils";
 
   interface Props {
-    onintersect: () => Promise<void>;
+    onIntersect: () => Promise<void>;
     layout?: "list" | "grid";
     disabled?: boolean;
     testId?: string;
@@ -14,7 +14,7 @@
   }
 
   let {
-    onintersect,
+    onIntersect,
     layout = "list",
     disabled = false,
     testId,
@@ -36,7 +36,7 @@
       return;
     }
 
-    await onintersect();
+    await onIntersect();
   };
 
   const observer: IntersectionObserver = new IntersectionObserver(
--- a/src/lib/components/InfiniteScroll.svelte
+++ b/src/lib/components/InfiniteScroll.svelte
@@ -1,26 +1,31 @@
 <script lang="ts">
-  import {
-    afterUpdate,
-    beforeUpdate,
-    createEventDispatcher,
-    onDestroy,
-  } from "svelte";
+  import { onDestroy, type Snippet } from "svelte";
   import { isNullish } from "@dfinity/utils";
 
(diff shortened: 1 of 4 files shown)


src/lib/components/InfiniteScroll.svelte
```diff
@@ -1,27 +1,32 @@
...
-    onDestroy,
-  } from "svelte";
+  import { onDestroy, type Snippet } from "svelte";
   import { isNullish } from "@dfinity/utils";
 
-  export let layout: "list" | "grid" = "list";
-  export let disabled = false;
-  export let testId: string | undefined = undefined;
+  interface Props {
+    onintersect: () => void;
```
peterpeterparker, line 6: We probably need a `Promise<void>` here, can you check OISY and NNS dapp?

Should we use `onintersect` or `onIntersect` - what's the current pattern in NNS dapp and OISY?

AntonioVentilii: you are right! both of them uses async functions for the event. Changed, have a look pls

src/lib/components/InfiniteScroll.svelte
```diff
@@ -1,28 +1,33 @@
...
-    onDestroy,
-  } from "svelte";
+  import { onDestroy, type Snippet } from "svelte";
   import { isNullish } from "@dfinity/utils";
 
-  export let layout: "list" | "grid" = "list";
-  export let disabled = false;
-  export let testId: string | undefined = undefined;
+  interface Props {
+    onIntersect: () => Promise<void>;
```
yhabib, line 6: This should probably be non-mandatory since `disabled=true`, no?

AntonioVentilii: ah! I thought we wanted it always passed on, just to guarantee that, if enabled, there is something to do

I can change it if you prefer

peterpeterparker: Which use case would use this component with "disable=true" and no callback?

peterpeterparker: Yeah, it sounds unexpected. Maybe it's just a placeholder because it might require an infinite scroller in the future. In that case, binding a callback with a TODO comment wouldn't be too bad anyway I'm thinking, but maybe there are other reasons. Do you remember @mstrasinskis?

PR #631, opened 2025-05-15
feat!: Migrate `Back` button to Svelte 5

AntonioVentilii:
# Motivation

Migrating component Back button to Svelte 5.

# Changes

- Migrate events in `Back` component, using `stopPropagation` wrapper.
- Adapt the usage in `Header` component: it will need to dispatch the old event `nnsBack`. We will remove it when we migrate it to Svelte 5.

# Screenshots

Current tests are sufficient.

### DIFF
--- a/src/lib/components/Back.svelte
+++ b/src/lib/components/Back.svelte
@@ -2,10 +2,10 @@
   import IconBack from "$lib/icons/IconBack.svelte";
   import { i18n } from "$lib/stores/i18n";
   import { stopPropagation } from "$lib/utils/event-modifiers.utils";
-  import type { GenericCallback } from "$lib/types/functions";
+  import type { OnEventCallback } from "$lib/types/event-modifiers";
 
   interface Props {
-    onBack: GenericCallback;
+    onBack: OnEventCallback;
   }
 
   let { onBack }: Props = $props();
--- a/src/lib/components/Back.svelte
+++ b/src/lib/components/Back.svelte
@@ -1,15 +1,20 @@
 <script lang="ts">
   import IconBack from "$lib/icons/IconBack.svelte";
   import { i18n } from "$lib/stores/i18n";
-  import { createEventDispatcher } from "svelte";
+  import { stopPropagation } from "$lib/utils/event-modifiers.utils";
+  import type { GenericCallback } from "$lib/types/functions";
 
-  const dispatch = createEventDispatcher();
+  interface Props {
+    onBack: GenericCallback;
+  }
+
+  let { onBack }: Props = $props();
 </script>
 
 <button
   class="back icon-only"
   data-tid="back"
-  on:click|stopPropagation={() => dispatch("nnsBack")}
+  onclick={stopPropagation(onBack)}
   aria-label={$i18n.core.back}><IconBack /></button
 >
 
--- a/src/lib/components/Header.svelte
+++ b/src/lib/components/Header.svelte
@@ -3,15 +3,22 @@
   import MenuButton from "$lib/components/MenuButton.svelte";
   import Back from "$lib/components/Back.svelte";
   import { layoutContentTopHidden } from "$lib/stores/layout.store";
+  import { createEventDispatcher } from "svelte";
 
   export let back = false;
+
+  const dispatch = createEventDispatcher();
+
+  const onBack = () => {
+    dispatch("nnsBack");
+  };
 </script>
 
 <header data-tid="header-component" class:hidden={$layoutContentTopHidden}>
   <Toolbar>
     <svelte:fragment slot="start">
       {#if back}
-        <Back on:nnsBack />
+        <Back {onBack} />
       {:else}
         <MenuButton />
       {/if}
--- a/src/lib/utils/event-modifiers.utils.ts
+++ b/src/lib/utils/event-modifiers.utils.ts
@@ -1,0 +1,10 @@
+import type { GenericCallback } from "$lib/types/functions";
+
+export const stopPropagation = (fn: GenericCallback) => {
+  return async (
+    event?: MouseEvent & { currentTarget: EventTarget & HTMLButtonElement },
+  ) => {
+    event?.stopPropagation();
+    await fn();
+  };
+};
--- a/src/lib/utils/event-modifiers.utils.ts
+++ b/src/lib/utils/event-modifiers.utils.ts
@@ -1,6 +1,9 @@
-// In Svelte 5 the events do not have modifiers anymore.
-// This module contains wrapper functions of event modifiers for Svelte 5.
-// Documentation: {@link https://svelte.dev/docs/svelte/v5-migration-guide#Event-changes-Event-modifiers}
+/**
+ * In Svelte 5, event modifiers are no longer part of the framework.
+ * This module contains wrapper functions that replicate modifier behavior for Svelte 5.
+ *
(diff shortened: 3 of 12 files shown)


src/lib/components/Back.svelte
```diff
@@ -1,15 +1,20 @@
 <script lang="ts">
   import IconBack from "$lib/icons/IconBack.svelte";
   import { i18n } from "$lib/stores/i18n";
-  import { createEventDispatcher } from "svelte";
+  import { stopPropagation } from "$lib/utils/event-modifiers.utils";
+  import type { GenericCallback } from "$lib/types/functions";
 
-  const dispatch = createEventDispatcher();
+  interface Props {
```
yhabib, line 7: Out of curiosity, do you prefer `interface` or `type`? In the nns-dapp, we chose `type` as the default and are wondering about the approaches of other teams. :)

src/lib/components/Back.svelte
```diff
@@ -1,15 +1,20 @@
...
+    onBack: GenericCallback;
+  }
+
+  let { onBack }: Props = $props();
```
yhabib, line 11: We also differentiated between `props` that the component modifies using `let` and those that it does not modify using `const`.

Not saying we have to do the same here, just sharing it :)

src/lib/components/Back.svelte
peterpeterparker: I use `interface`.

src/lib/components/Back.svelte
peterpeterparker: We discussed this recently in a PR with @mstrasinskis.  It seems to be a matter of personal preference.

As for me, since the variable used to have to be declared with `let` and considering that Svelte itself uses `let` in its documentation and examples, I tend to prefer sticking with `let`.

Additionally, in discussions on the Svelte repo — [here](https://github.com/sveltejs/svelte/issues/12121) and [here](https://github.com/sveltejs/eslint-plugin-svelte/issues/818) — one developer mentioned that `$bindable` must be declared with `let`. I’m not sure if that’s accurate as I haven’t tested it myself, but if it is, avoiding differences would be another good reason to stick with `let`.

That said, again, it’s probably just a matter of preference.

src/lib/components/Back.svelte
peterpeterparker: Doesn't our [linter](https://github.com/dfinity/eslint-config-oisy-wallet) also inforces interfaces maybe?

AntonioVentilii: Yes, there is a rule for it (I am actually fixing it with https://github.com/dfinity/eslint-config-oisy-wallet/pull/114), but we are not using that ESlint package here. I can introduce it, if it is something we can agree on

src/lib/components/Back.svelte
AntonioVentilii: On my side: `let`. I agree with @peterpeterparker, always preferable to follow the official documentation. Plus, especially for bindings (since it seems it's not clear), we will probably use `let`, and it becomes a bit inconsistent.

AntonioVentilii: I tested what @peterpeterparker mentioned about bindable. Indeed, with `const` you cannot set it as `bindable`. However, it was a good warning to find out which props MUST be `bindable`... Because I was warned that I was changing them.

src/lib/components/Back.svelte
peterpeterparker: I started a [draft](https://github.com/dfinity/gix-components/pull/590) in February to integrate the ESLint rules into this library. I didn't have time to come back to it, so I'll close the PR and let you take over.

Regarding interfaces vs. types, I like both and I don't have a strong opinion as long as it's consistent. Therefore, since the library currently uses `interface` for its `types` and `stores` packages — aside from one inconsistency— and the ESLint rules enforce `interface` as well, I would go with `interface`. It seems the more consistent path.

AntonioVentilii: One for `interface`

yhabib: Sounds reasonable for me

src/lib/components/Header.svelte
```diff
@@ -3,15 +3,22 @@
   import MenuButton from "$lib/components/MenuButton.svelte";
   import Back from "$lib/components/Back.svelte";
   import { layoutContentTopHidden } from "$lib/stores/layout.store";
+  import { createEventDispatcher } from "svelte";
 
   export let back = false;
```
yhabib, line 8: I guess this will be migrated in a different PR.

yhabib (approved):

src/lib/utils/event-modifiers.utils.ts
```diff
@@ -0,0 +1,10 @@
+import type { GenericCallback } from "$lib/types/functions";
```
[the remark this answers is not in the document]
AntonioVentilii: - you mean `GenericCallback`? It is in `$lib/types`; you mean another `types` folder?
- yes, agreed: what if `OnEventFunction` or `OnEventCallback`?

src/lib/utils/event-modifiers.utils.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export const stopPropagation = (fn: GenericCallback) => {
```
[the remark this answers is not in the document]
AntonioVentilii: Done with:

- https://github.com/dfinity/gix-components/pull/640

src/lib/utils/event-modifiers.utils.ts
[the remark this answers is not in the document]
peterpeterparker: - In `$lib/types`
- Maybe `OnClickEvent`?

src/lib/utils/event-modifiers.utils.ts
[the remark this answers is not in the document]
AntonioVentilii: included the return type in the dedicated PR

src/lib/utils/event-modifiers.utils.ts
```diff
@@ -0,0 +1,10 @@
...
+  return async (
+    event?: MouseEvent & { currentTarget: EventTarget & HTMLButtonElement },
```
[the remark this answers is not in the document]
AntonioVentilii: agreed!

AntonioVentilii:
@yhabib @peterpeterparker ready for another review pls

src/tests/lib/components/Back.spec.ts
```diff
@@ -6,8 +6,8 @@ describe("Back", () => {
   it("should forward the click event", () =>
     new Promise<void>((done) => {
       const { getByTestId } = render(Back, {
-        events: {
-          nnsBack: () => done(),
+        props: {
+          onBack: () => done(),
```
peterpeterparker, line 10: if you like it, can be shortened: `onBack: done`

peterpeterparker (approved):
LGTM, thx 👍

You can edit the description of the PR, the change "Create util to replicate even..." can notably be removed now.

src/lib/utils/event-modifiers.utils.ts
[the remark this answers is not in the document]
peterpeterparker: Resolved in separate PR

src/tests/lib/components/Back.spec.ts
AntonioVentilii: done!

yhabib:
LGTM , thanks

PR #634, opened 2025-05-15
feat: Migrate component `Html` to Svelte 5

AntonioVentilii:
# Motivation

Migrating component `Html` to Svelte 5.


peterpeterparker (approved):
LGTM, thx

If someone has time, maybe we could review if the mount rerender is still required but, not related to the migration.

PR #638, opened 2025-05-15
feat: Migrate component `Markdown` to Svelte 5

AntonioVentilii:
# Motivation

Migrating component `Markdown` to Svelte 5.

### DIFF
--- a/src/lib/components/Markdown.svelte
+++ b/src/lib/components/Markdown.svelte
@@ -3,10 +3,14 @@
   import Spinner from "$lib/components/Spinner.svelte";
   import Html from "$lib/components/Html.svelte";
 
-  export let text: string | undefined;
+  interface Props {
+    text: string | undefined;
+  }
 
-  let html: string | undefined;
-  let error = false;
+  let { text }: Props = $props();
+
+  let html: string | undefined = $state();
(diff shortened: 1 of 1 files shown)


src/lib/components/Markdown.svelte
```diff
@@ -3,10 +3,14 @@
   import Spinner from "$lib/components/Spinner.svelte";
   import Html from "$lib/components/Html.svelte";
 
-  export let text: string | undefined;
+  interface Props {
+    text: string | undefined;
```
AntonioVentilii, line 7: the previous behaviour is to provide the text even if undefined. I kept the same behaviour

src/lib/components/Markdown.svelte
```diff
@@ -3,10 +3,14 @@
...
+  }
 
-  let html: string | undefined;
-  let error = false;
+  let { text }: Props = $props();
+
+  let html: string | undefined = $state();
```
peterpeterparker, line 12:
```
let html = $state<string | undefined>()
```

maybe?

AntonioVentilii: done!

peterpeterparker (approved):
LGTM, thx

PR #640, opened 2025-05-17
feat: Util for event modifier `stopPropagation`

AntonioVentilii:
# Motivation

According to [Svelte 5 documentation](https://svelte.dev/docs/svelte/v5-migration-guide#Event-changes-Event-modifiers), with the migration, the event modifiers are deprecated and should be replaced by functions that have the same scope. For example, instead of:

```svelte
<button on:click|preventDefault={handler}>...</button>
```

One should do

```svelte
<script>
	function preventDefault(fn) {
		return function (event) {
			event.preventDefault();
			fn.call(this, event);
		};
	}
</script>

<button onclick={once(preventDefault(handler))}>...</button>
```

In this PR we create the `stopPropagation` util, to help with the migration (it will be used in other PR, like https://github.com/dfinity/gix-components/pull/631).

# Changes

- Creating a new type for event handlers.
- Creating new util for stop propagation, based on the Svelte documentation.
- Export new event-modifiers util modules.

# Screenshots

Created new tests, including a mock component.

### DIFF
--- a/src/lib/types/event-modifiers.ts
+++ b/src/lib/types/event-modifiers.ts
@@ -1,5 +1,1 @@
 export type OnEventCallback = () => void | Promise<void>;
-
-export type OnMouseEventHandler = (
-  $event?: MouseEvent & { currentTarget: EventTarget & HTMLButtonElement },
-) => Promise<void>;
--- a/src/lib/types/event-modifiers.ts
+++ b/src/lib/types/event-modifiers.ts
@@ -1,5 +1,5 @@
 export type OnEventCallback = () => void | Promise<void>;
 
 export type OnMouseEventHandler = (
-  event?: MouseEvent & { currentTarget: EventTarget & HTMLButtonElement },
+  $event?: MouseEvent & { currentTarget: EventTarget & HTMLButtonElement },
 ) => Promise<void>;
--- a/src/lib/types/event-modifiers.ts
+++ b/src/lib/types/event-modifiers.ts
@@ -1,0 +1,5 @@
+export type OnEventCallback = () => void | Promise<void>;
+
+export type OnMouseEventHandler = (
+  event?: MouseEvent & { currentTarget: EventTarget & HTMLButtonElement },
+) => Promise<void>;
--- a/src/lib/utils/event-modifiers.utils.ts
+++ b/src/lib/utils/event-modifiers.utils.ts
@@ -1,3 +1,7 @@
+// In Svelte 5 the events do not have modifiers anymore.
+// This module contains wrapper functions of event modifiers for Svelte 5.
+// Documentation: https://svelte.dev/docs/svelte/v5-migration-guide#Event-changes-Event-modifiers
+
 import type { OnEventCallback } from "$lib/types/event-modifiers";
 import type { MouseEventHandler } from "svelte/elements";
 
--- a/src/lib/utils/event-modifiers.utils.ts
+++ b/src/lib/utils/event-modifiers.utils.ts
@@ -8,6 +8,13 @@
 import type { OnEventCallback } from "$lib/types/event-modifiers";
 import type { MouseEventHandler } from "svelte/elements";
 
+/**
+ * A wrapper function to stop event propagation of a mouse event before executing a callback function.
+ *
+ * @param {OnEventCallback} fn - The function to be executed after stopping the event propagation. It can be a synchronous or asynchronous function.
+ *
+ * @returns {MouseEventHandler<T extends EventTarget>} - A function that takes an event and stop its propagation, before executing the provided function.
+ */
 export const stopPropagation = <T extends EventTarget>(
   fn: OnEventCallback,
 ): MouseEventHandler<T> => {
--- a/src/lib/utils/event-modifiers.utils.ts
+++ b/src/lib/utils/event-modifiers.utils.ts
@@ -10,11 +10,11 @@ import type { MouseEventHandler } from "svelte/elements";
  *
  * @param {OnEventCallback} fn - The function to be executed after stopping the event propagation. It can be a synchronous or asynchronous function.
  *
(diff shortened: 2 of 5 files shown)


src/lib/types/event-modifiers.ts
```diff
@@ -0,0 +1,5 @@
+export type OnEventCallback = () => void | Promise<void>;
+
+export type OnMouseEventHandler = (
+  $event?: MouseEvent & { currentTarget: EventTarget & HTMLButtonElement },
```
peterpeterparker, line 4: A detail but, I wonder what's the correct type because the click event can be triggered by a touch or keyboard event as well. I guess MouseEvent is fine but, technically speaking it can be something else.

src/lib/types/event-modifiers.ts
```diff
@@ -0,0 +1,5 @@
...
+) => Promise<void>;
```
peterpeterparker, line 5: Should it also accept `void`? `Promise<void> | void`? Otherwise we limit the usage to promises only but, I guess there will be synchronous usage like when a button just update a state.

Or we create two handlers? `OnSomethingEvent` and `OnSomethingEventAsync`?

src/lib/utils/event-modifiers.utils.ts
```diff
@@ -0,0 +1,13 @@
+import type {
+  OnEventCallback,
+  OnMouseEventHandler,
+} from "$lib/types/event-modifiers";
+
+export const stopPropagation = (fn: OnEventCallback): OnMouseEventHandler => {
```
peterpeterparker, line 6: It does not just stop the propagation, it also runs the callback. We probably need a better naming here.

src/lib/types/event-modifiers.ts
AntonioVentilii: the specific return handler must be async, because we pass it as

const handler = eventModifier(ourFunction)

onclick={handler}

so, handler could be whatever, `void` or `Promise<void>`; however, since `ourFunction` can be both, we are forced to `await` it inside `handler`, making `handler` compulsory async.

src/lib/types/event-modifiers.ts
AntonioVentilii: right, good point! let me find out if there is a more generic type

src/lib/utils/event-modifiers.utils.ts
AntonioVentilii: i was following the Svelte docs, where they keep the same name as the previous event modifier, i imagine for continuity. But we can rename it as `stopPropagationAndRun`?

src/lib/types/event-modifiers.ts
peterpeterparker: Oki doki

src/lib/types/event-modifiers.ts
AntonioVentilii: Strange.. in Svelte lib, the `onclick` is defined as `MouseEvent`:

```ts
export interface DOMAttributes<T extends EventTarget> {
  ...
  // MouseEvents
  'on:auxclick'?: MouseEventHandler<T> | undefined | null;
  onauxclick?: MouseEventHandler<T> | undefined | null;
  onauxclickcapture?: MouseEventHandler<T> | undefined | null;
  'on:click'?: MouseEventHandler<T> | undefined | null;
  onclick?: MouseEventHandler<T> | undefined | null;  <<<-------------------here
  onclickcapture?: MouseEventHandler<T> | undefined | null;
  'on:contextmenu'?: MouseEventHandler<T> | undefined | null;
  oncontextmenu?: MouseEventHandler<T> | undefined | null;
  oncontextmenucapture?: MouseEventHandler<T> | undefined | null;
  ...
}
```

But, I am now using the same types of Svelte lib

src/lib/utils/event-modifiers.utils.ts
peterpeterparker: Ah I see, it's good then, let's stick to `stopPropagation`.

Can you however add some JSDocs and/or the link to the doc then?

AntonioVentilii: for sure

src/lib/types/event-modifiers.ts
peterpeterparker: nice 👍

src/lib/utils/event-modifiers.utils.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * In Svelte 5 the events do not have modifiers anymore.
```
peterpeterparker: I would document it regardless of Svelte v5 - i.e. describe what it does

src/lib/utils/event-modifiers.utils.ts
```diff
@@ -0,0 +1,18 @@
...
+ * This module contains wrapper functions of event modifiers for Svelte 5.
+ *
+ * Documentation: https://svelte.dev/docs/svelte/v5-migration-guide#Event-changes-Event-modifiers
```
peterpeterparker: `@link`?

peterpeterparker (approved):
LGTM, thx

I guess we will start using those utils in our dapps as well, maybe we can think about adding a utilities section in the documation. Just thinking out at loud, not a call to adction.

src/lib/utils/event-modifiers.utils.ts
AntonioVentilii: done!

src/lib/utils/event-modifiers.utils.ts
AntonioVentilii: included

peterpeterparker:
> ust thinking out at loud, not a call to adction.

I gave the doc a try: https://github.com/dfinity/gix-components/pull/642

PR #642, opened 2025-05-17
feat: auto-generate docs for utilities

peterpeterparker:
# Motivation

Today we added a utility in #640 that we are exposing because it can be useful for consumer as well. Hence, I thought it would be nice to automatically generate the documentation.

# Changes

-`tsdoc-markdown` **
  - Install dependency
  - Create script to generate docs (target +page.md instead of a README.md)
  - Add a script in package.json
  - Add GitHub Actions to generate docs (scope it to only utils and page)
- Add new route '/utility-functions' and related menu entry

** Same setup as in IC-JS repo.

# Notes

The screenshots need to be regenerated because we are modifying the menu with this PR.

# Screenshots

Capture d’écran 2025-05-17 à 13 47 48


.github/actions/prepare/action.yml
```diff
@@ -0,0 +1,14 @@
+name: Prepare
```
AntonioVentilii, line 1: maybe this one can be done in a separate PR? since I see that it can be used for the deploy workflow too (and I imagine some more?). Or are you planning of doing it after?

peterpeterparker: I just copied the scripts 1 to 1 from from ic-js, that's why I did not modified it how it works: https://github.com/dfinity/ic-js/tree/main/.github

But happy to split it, what do you prefer, should I provide the actions in a separate PR? Inline the scripts? etc. let me know

AntonioVentilii: No strong opinion honestly: we can proceed with this one and then in another PR use the same action for the other workflows (where necessary)

AntonioVentilii (approved):
Good idea! LGTM tks!

PR #643, opened 2025-05-18
chore: use CI prepare action

peterpeterparker:
# Motivation

We want to use the "prepare" action recently introduced (cloned from ic-js and other repo) to setup Node LTS and install dependencies in every CI jobs.


AntonioVentilii (approved):
LGTM tks!

PR #652, opened 2025-05-26
feat(icons): add check-circle icon v2

yhabib:
# Motivation

The nns-dapp requires a specific set of icons to be displayed together for a new feature. Gix already has an icon called `IconCheckCircle`, but it is incompatible with the other icons due to its creation method.

This PR introduces version 2 of this icon. The plan is to replace the original icon in a later step to avoid having multiple versions.

Screenshot 2025-05-26 at 13 19 42

# Changes

- Adds `IconCheckCircleV2`icon
- Cleans up `IconCheckCircleFill` with [svgomg](https://jakearchibald.github.io/svgomg/).
- Updates `IconCloseCircleFill`

# Screenshots

Rendering side by side in the nns-dapp:

Screenshot 2025-05-26 at 13 41 05


mstrasinskis (approved):
Thanks!

PR #657, opened 2025-06-27
feat: Migrate component `Island` to Svelte 5

AntonioVentilii:
# Motivation

Migrating component `Island` to Svelte 5.


src/lib/components/Island.svelte
```diff
@@ -3,24 +3,32 @@
   import { layoutContentScrollY } from "$lib/stores/layout.store";
   import { BREAKPOINT_LARGE } from "$lib/constants/constants";
   import ScrollSentinel from "$lib/components/ScrollSentinel.svelte";
+  import type { Snippet } from "svelte";
 
-  export let testId: string | undefined = undefined;
+  interface Props {
+    testId?: string;
+    children?: Snippet;
```
peterpeterparker, line 10: Mandatory?

src/lib/components/Island.svelte
```diff
@@ -3,24 +3,32 @@
...
-  let innerWidth = 0;
-  $: innerWidth,
+  $effect(() => {
     layoutContentScrollY.set(innerWidth < BREAKPOINT_LARGE ? "auto" : "hidden");
+  });
 
   onDestroy(() => layoutContentScrollY.set("auto"));
 
-  let scrollContainer: HTMLElement;
+  let scrollContainer = $state<HTMLElement | undefined>();
```
peterpeterparker, line 23: `ScrollSentinel` seems to expect a defined `scrollContainer`, should this component be migrated as well?

AntonioVentilii: don't we prefer that `ScrollSentinel` takes in a non-nullish value? I can add a check for nullish values directly here in the `Island` component

peterpeterparker: I would not change the behavior but only adapt the types - we are migrating to Svelte v5, not changing the JS feature. Later on why not. You know what I mean?

PR #662, opened 2025-07-04
feat: Expose i18n store

AntonioVentilii:
# Motivation

We would like to add more languages to the localization of GIX Components when using it in OISY.

As first solution, we expose i18n store, so that the consumer app (in our case OISY) can set the language when setting it for the rest too.


peterpeterparker (changes requested):
This does not seems correct. The language should automatically switch and therefore the store does not need to be exposed for that particular feature..

When it comes to supported language, I would rather add the translation to the library rather than implementing custom languages per dapps. This seems like a standard approach for libraries that supports i18n.

PR #663, opened 2025-07-07
feat: Migrate component `ScrollSentinel` to Svelte 5

AntonioVentilii:
# Motivation

Migrating component `ScrollSentinel` to Svelte 5.

However, Svelte 5 is stricter on the `$state` types, meaning that internal variable element could be `undefined` (not only `HTMLElement`).

To adapt to this change, we need to refactor a bit the `onMount` logic: we should await for the DOM and the binding to finish through the Svelte service `tick`.

### DIFF
--- a/src/lib/components/Copy.svelte
+++ b/src/lib/components/Copy.svelte
@@ -1,8 +1,16 @@
 <script lang="ts">
   import { IconCopy } from "$lib/icons";
   import { i18n } from "$lib/stores/i18n";
+  import {
+    stopPropagation,
+    preventDefault,
+  } from "$lib/utils/event-modifiers.utils";
 
-  export let value: string;
+  interface Props {
+    value: string;
+  }
+
+  let { value }: Props = $props();
 
   const copyToClipboard = async () =>
     await navigator.clipboard.writeText(value);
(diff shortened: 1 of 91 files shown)


src/lib/components/ScrollSentinel.svelte
```diff
@@ -1,21 +1,41 @@
...
-    const observer = new IntersectionObserver(
-      ([entry]) => layoutContentTopHidden.set(!entry.isIntersecting),
-      { root: scrollContainer, threshold: 0 },
-    );
-    observer.observe(element);
-    return () => observer.disconnect();
+    let observer: IntersectionObserver | undefined = undefined;
+
+    const setObserver = async () => {
+      await tick(); // wait for DOM + bind:this to complete
```
peterpeterparker, line 22: Mmm that's recommended by the doc? In `InfiniteScroll` with use an `$effect`, would that work here as well?

peterpeterparker: Kind of, like in InfiniteScoll https://github.com/dfinity/gix-components/blob/3f28aa0b44ee668ea143e1928d518def7401eebb/src/lib/components/InfiniteScroll.svelte#L42C3-L45C5

```
const observer: IntersectionObserver = new IntersectionObserver(
    onIntersection,
    options,
  );
  
  $effect(() =>. {
     observer.unsubribe
      
      observer.subscribe
  
  )
  
  ondestory(unsubscribe)
  ```

peterpeterparker: Looks great, thanks!

PR #667, opened 2025-07-07
feat: Migrate component `TestIdWrapper` to Svelte 5

AntonioVentilii:
# Motivation

Migrating component `TestIdWrapper` to Svelte 5.

### DIFF
--- a/src/lib/components/TestIdWrapper.svelte
+++ b/src/lib/components/TestIdWrapper.svelte
@@ -1,9 +1,16 @@
 <script lang="ts">
-  export let testId: string | undefined = undefined;
+  import type { Snippet } from "svelte";
+
+  interface Props {
+    testId: string;
+    children: Snippet;
+  }
+
+  let { testId, children }: Props = $props();
 </script>
 
 <div class="contents" data-tid={testId}>
-  <slot />
+  {@render children()}
 </div>
 
 <style lang="scss">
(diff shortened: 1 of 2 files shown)


src/lib/components/TestIdWrapper.svelte
```diff
@@ -1,9 +1,16 @@
 <script lang="ts">
-  export let testId: string | undefined = undefined;
+  import type { Snippet } from "svelte";
+
+  interface Props {
+    testId: string;
```
peterpeterparker, line 5: It's fine to mark it as mandatory, no usage in NNS-dapp with the value for some reasons set to undefined?

AntonioVentilii: nns-dapp is using its own `TestIdWrapper`, but in any case it is compulsory there too. This is the code of `TestIdWrapper` in nns-dapp:

```svelte
<script lang="ts">
  export let testId: string;
</script>

<div class="contents" data-tid={testId}>
  <slot />
</div>

<style lang="scss">
  .contents {
    display: contents;
  }
</style>
```


ref https://github.com/dfinity/nns-dapp/blob/c042a760aa33b552a52ba5bdd32ff200745ab555/frontend/src/lib/components/common/TestIdWrapper.svelte

peterpeterparker: Thanks for double checking!

PR #675, opened 2025-07-07
fix: Pass event to callback of event modifiers

AntonioVentilii:
# Motivation

We should pass the event to the event modifiers callback. That is what is suggested in the official documentation too (see https://svelte.dev/docs/svelte/v5-migration-guide#Event-changes-Event-modifiers), but in general, it makes sense. Especially, if we use the modifiers concatenated.

# Changes

- Create simplified type for event parameter.
- Pass the event to the call back function in both exiting modifiers.
- Removing optionality of events, since it is not necessary.

# Screenshots

Current tests were adapted to test this new change. The unchanged tests must work as usual.


peterpeterparker (approved):
LGTM, thx

PR #677, opened 2025-07-07
feat: Define CSS classes for levels of `Toast` component

AntonioVentilii:
# Motivation

Just an idea: it would be easier to customize the Toast style using CSS styles, if we add the level among the classes.

For example, in OISY, if we want to customize the background of an error toast we need to use the icon child:

```css
.toast:has(> .icon.error) {
	--overlay-background: var(--color-background-error-light);
	--overlay-background-contrast: var(--color-foreground-error-secondary);
}
```

While we could do:

```css
.toast {
	--overlay-background-error: var(--color-background-error-light);
	--overlay-background-contrast-error: var(--color-foreground-error-secondary);
}
```

So, we define custom CSS classes for each level of `Toast`: background and foreground, inverted or not.

### DIFF
--- a/src/lib/components/InputRange.svelte
+++ b/src/lib/components/InputRange.svelte
@@ -1,12 +1,21 @@
 <script lang="ts">
-  export let max: number;
-  export let min: number;
-  export let value: number;
-  export let ariaLabel: string;
-  export let handleInput: (() => void) | undefined = undefined;
+  interface Props {
+    max: number;
+    min: number;
+    value: number;
+    ariaLabel: string;
+    handleInput?: (() => void) | undefined;
+  }
 
-  let progression: number;
-  $: progression = Math.round(((value - min) / (max - min)) * 100);
+  let {
+    max,
+    min,
+    value = $bindable<number>(),
+    ariaLabel,
+    handleInput = undefined,
+  }: Props = $props();
+
+  let progression = $derived(Math.round(((value - min) / (max - min)) * 100));
 </script>
 
 <!-- Order of on:input and bind:value matters: https://svelte.dev/docs#template-syntax-element-directives-bind-property -->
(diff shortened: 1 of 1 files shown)


peterpeterparker:
Or CSS variables? 

Something like 

```
.toast{
	--overlay-background-error: var(--color-background-error-light);
}
```

Not sure what's the common pattern but, I guess we rarely assume the consumer uses CSS selector, we probably do so in OISY as a hack...

AntonioVentilii:
@peterpeterparker you mean creating an `--overlay-background-*` for each `level`, right? That would be coherent with how we normally use GIX-cmp

peterpeterparker:
I was thinking defining those in the `<style />` tag based on the `class` selector but, yes.

peterpeterparker (commented):
Wait so it was already there, we just want to have more flexibility, is that right?

If so, can you update title and description of the PR accordingly?

src/lib/components/Toast.svelte
```diff
@@ -140,8 +140,62 @@
     padding: var(--padding-1_5x);
     box-sizing: border-box;
 
+    background: var(--toast-background, var(--overlay-background));
+    color: var(--toast-background-contrast, var(--overlay-background-contrast));
+
+    &.success {
+      --toast-background: var(--toast-background-success);
+      --toast-background-contrast: var(--toast-background-contrast-success);
```
peterpeterparker, line 148: I wonder if we should not name it `--toast-color`? The variable is about the toast not the color pattern, wdyt?

src/lib/components/Toast.svelte
```diff
@@ -140,8 +140,62 @@
...
+    }
+
+    &.info {
+      --toast-background: var(--toast-background-info);
```
peterpeterparker, line 152: For consistency, should we invert the "background" and "info" because we have existing variables which are named such as `--toast-info-z-index` or `--toast-error-z-index`?

src/lib/components/Toast.svelte
AntonioVentilii: yes, makes sense, eheh! I was in doubt myself

peterpeterparker (approved):
LGTM, thx

PR #678, opened 2025-07-08
fix: Avoid loss of precision in Input component

AntonioVentilii:
# Motivation

The Input component uses `Number()` when it is a currency type. However, due to floating-point limitations of this method, the value that the user inputs is different from the rendered ones sometimes. For example:

Screenshot 2025-07-08 at 16 19 13

Unfortunately there is no real solution unless we create our own `Number(value).toLocaleString()` method. Or if we use a common library like [decimals.js](https://www.npmjs.com/package/decimal.js?activeTab=readme).

We will opt for the second one, with a caveat: `Decimals` raises an error for empty strings, or non-number strings, while `Number` would just create the object `NaN`.

So, we make a safe-parsing of `Decimals`, and fallback to the "old" method if it fails.

# Changes

- Install `decimals.js`.
- Change Input component to use Decimals but in a safe way (as explained above).

# Screenshots

Added tests and tested manually:

Screenshot 2025-07-08 at 16 24 00

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "vitest": "^3.1.1"
   },
   "dependencies": {
+    "decimal.js": "^10.6.0",
     "dompurify": "^3.2.4",
     "html5-qrcode": "^2.3.8",
     "marked": "^9.1.0",
--- a/src/lib/components/Input.svelte
+++ b/src/lib/components/Input.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
   import { isNullish, nonNullish } from "@dfinity/utils";
   import { createEventDispatcher } from "svelte";
+  import Decimal from "decimal.js";
 
   export let name: string;
   export let inputType: "icp" | "number" | "text" | "currency" = "number";
@@ -42,11 +43,26 @@
   let selectionStart: number | null = 0;
   let selectionEnd: number | null = 0;
 
-  const toStringWrapDecimals = (value: string): string =>
-    Number(value).toLocaleString("en", {
-      useGrouping: false,
-      maximumFractionDigits: wrapDecimals,
-    });
+  const safeDecimal = (value: unknown): Decimal | null => {
+    try {
+      return new Decimal(value as string);
+    } catch {
+      return null;
+    }
+  };
+
+  const toStringWrapDecimals = (value: string): string => {
+    const decimalValue = safeDecimal(value);
+
+    if (isNullish(decimalValue)) {
+      return Number(value).toLocaleString("en", {
+        useGrouping: false,
+        maximumFractionDigits: wrapDecimals,
+      });
+    }
+
+    return decimalValue.toDecimalPlaces(wrapDecimals).toFixed();
+  };
 
   // replace exponent format (1e-4) w/ plain (0.0001)
   const exponentToPlainNumberString = (value: string): string =>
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -74,6 +74,7 @@
     "vitest": "^3.1.1"
   },
   "dependencies": {
+    "decimal.js": "^10.6.0",
```
peterpeterparker, line 77: Obviously, no when it comes to me ;)

peterpeterparker (changes requested):
> The Input component uses Number() when it is a currency type.

This is innacurate. For icp and currency the input type is of type text. If there is a rounding issue, it means there is a glitch in the post processing of the inputs.

Have you identified the exact root cause of the issue and when does the amount looses its precision?

src/lib/components/Input.svelte
```diff
@@ -42,11 +43,26 @@
   let selectionStart: number | null = 0;
   let selectionEnd: number | null = 0;
 
-  const toStringWrapDecimals = (value: string): string =>
-    Number(value).toLocaleString("en", {
-      useGrouping: false,
-      maximumFractionDigits: wrapDecimals,
-    });
+  const safeDecimal = (value: unknown): Decimal | null => {
```
peterpeterparker, line 46: Can you add some jsdocs on why we use this third party library?

src/lib/components/Input.svelte
```diff
@@ -42,11 +43,26 @@
...
+    } catch {
+      return null;
+    }
+  };
+
+  const toStringWrapDecimals = (value: string): string => {
+    const decimalValue = safeDecimal(value);
+
+    if (isNullish(decimalValue)) {
+      return Number(value).toLocaleString("en", {
```
peterpeterparker, line 58: Maybe a comment on why it's safe to fallback here

package.json
peterpeterparker: 🥲

src/tests/lib/components/Input.svelte.spec.ts
```diff
@@ -517,6 +518,32 @@ describe("Input", () => {
       expect(input.value).toBe("0.000000011231232121");
     });
 
+    it("should not trim below given decimals", async () => {
```
peterpeterparker, line 521: Can you add `0.999999999999999876` to the tests since you mentionned it as comment and I used it for having a look?

peterpeterparker (commented):
I guess no choice...

Thanks for the explanation

src/lib/components/Input.svelte
AntonioVentilii: Added, maybe a little verbose, but WDYT?

src/lib/components/Input.svelte
AntonioVentilii: Same as comment above

src/lib/components/Input.svelte
peterpeterparker: > * The native `Number()` can produce inaccurate results for such inputs. For example:
   * `Number("0.999999999999999876")` may yield a different value than what the user actually typed.
   
   
I would document the example that you shared to enhance clarity.

src/lib/components/Input.svelte
AntonioVentilii: updated: WDYT?

src/lib/components/Input.svelte
AntonioVentilii: done, WDYT?

peterpeterparker (approved):
LGTM, thx

PR #686, opened 2025-07-11
fix: Default testId for `Collapsible` component

AntonioVentilii:
# Motivation

The `Collapsible` component renders `TestIdWrapper`. This last component required a non-nullish `testId` (for obvious reasons). So, since this seems to be the intention of the code, we default the `testId` to a value.


peterpeterparker (approved):
LGTM, thx

Makes me think that - someday in another PR -we should prefix the default testId to avoid clashes. Like `gix-cmp-collapsible` instead of `collapsible`. But again, someday and generally speaking.

PR #690, opened 2025-07-11
chore: Add prefix to default test IDs

AntonioVentilii:
# Motivation

As suggested by @peterpeterparker in https://github.com/dfinity/gix-components/pull/686#pullrequestreview-3008842729, we add a prefix to all the default values of test IDs, to avoid possible conflicts with the consumers.


peterpeterparker (approved):
LGTM, thx!

This might break the test suite of NNS dapp and OISY so, once we merge it, we have to generate a next release and upgrade the consumers.

PR #689, opened 2025-07-11
feat!: Migrate `WizardModal` to Svelte 5

AntonioVentilii:
# Motivation

Migrating component `WizardModal` to Svelte 5.

# Breaking Changes

The interface changes:

- The slot `title` is migrated to `Snippet`.
- No event `nnsClose` bubbling up anymore, use the prop `onClose` that receives the callback function.

PR #695, opened 2025-07-11
fix: Component `WizardModal` is back to correct working

AntonioVentilii:
# Motivation

After migrating `WizardModal` to Svelte 5 in https://github.com/dfinity/gix-components/pull/689 , it is not working anymore using function `modal.next()`, `modal.back()` and `modal.set()`.

https://github.com/user-attachments/assets/ea0854ae-b0e5-486e-bcb3-d8605c1bcaf3

That is because the reactivity of `$effect` works different in the following snippet:

```svelte
// Svelte v4
$: ({ currentStep } = stepState)

// Svelte v5
$effect(() => {
  ({ currentStep } = stepState);
});
```
  
In Svelte 5, this reactivity is in fact triggered only when the entire reference of `stepState` changes. However, if only its properties change (like when we call functions `next`, `back` and `set`), then it does not trigger a reaction, causing a stale component.

A possible solution is to directly trigger the update of `currentStep` in each function.

We keep the $effect rune, just because it should be triggered on assigning any value to `stepState` in any case.

# Changes

- Create small function to update the `currentStep` variable, taking `stepState` as input.
- Use it both in the $effect rune and in all the exported functions: `next`, `back`, `set`.

# Screenshots

I created new tests (with a test component). They work now, but it can be easily verified that they were not working before this PR.

At the same here is a local manual test:

https://github.com/user-attachments/assets/072695c4-8219-44f9-99ce-886327592eb0

### DIFF
--- a/src/lib/components/WizardModal.svelte
+++ b/src/lib/components/WizardModal.svelte
@@ -3,7 +3,7 @@
 </script>
 
 <script lang="ts" generics="T extends string">
-  import type { Snippet } from "svelte";
+  import { type Snippet } from "svelte";
   import WizardTransition from "./WizardTransition.svelte";
   import Modal from "$lib/components/Modal.svelte";
   import { WizardStepsState } from "$lib/stores/wizard.state";
@@ -33,13 +33,25 @@
 
   let transition = $derived({ diff: stepState.diff });
 
-  $effect(() => {
+  const updateCurrentStep = (stepState: WizardStepsState<T>) =>
     ({ currentStep } = stepState);
+
+  $effect(() => {
+    updateCurrentStep(stepState);
   });
 
-  export const next = () => (stepState = stepState.next());
-  export const back = () => (stepState = stepState.back());
-  export const set = (step: number) => (stepState = stepState.set(step));
+  export const next = () => {
+    stepState = stepState.next();
+    updateCurrentStep(stepState);
+  };
+  export const back = () => {
+    stepState = stepState.back();
+    updateCurrentStep(stepState);
+  };
+  export const set = (step: number) => {
+    stepState = stepState.set(step);
+    updateCurrentStep(stepState);
+  };
 
   // onDestroy is not always called when repetitively opened/closed in NNS-dapp.
   // This might be linked to Svelte issue https://github.com/sveltejs/svelte/issues/5268.
--- a/src/tests/lib/components/WizardModal.spec.ts
+++ b/src/tests/lib/components/WizardModal.spec.ts
@@ -1,8 +1,9 @@
 import WizardModal from "$lib/components/WizardModal.svelte";
 import type { WizardStep, WizardSteps } from "$lib/types/wizard";
-import { render } from "@testing-library/svelte";
+import { fireEvent, render, waitFor } from "@testing-library/svelte";
 import type { Snippet } from "svelte";
 import { mockSnippet } from "../mocks/snippet.mocks";
+import WizardModalTest from "./WizardModalTest.svelte";
 
 describe("WizardModal", () => {
   const steps: WizardSteps = [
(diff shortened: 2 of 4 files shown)


src/lib/components/WizardModal.svelte
```diff
@@ -33,13 +33,25 @@
 
   let transition = $derived({ diff: stepState.diff });
 
-  $effect(() => {
+  const updateCurrentStep = (stepState: WizardStepsState<T>) =>
     ({ currentStep } = stepState);
+
+  $effect(() => {
+    updateCurrentStep(stepState);
```
peterpeterparker, line 40: This effect is fired right?

src/lib/components/WizardModal.svelte
```diff
@@ -33,13 +33,25 @@
...
   });
 
-  export const next = () => (stepState = stepState.next());
-  export const back = () => (stepState = stepState.back());
-  export const set = (step: number) => (stepState = stepState.set(step));
+  export const next = () => {
```
peterpeterparker, line 43: Maybe opinionated, not sure, but I feel like it would be cleaner to not duplicate `stepState = ...` in each function.

Maybe pass a callback?

```
const updateCurrentStep = (go: () => StepState) => {
   stepState = go();
   ({ currentStep } = stepState);
}

export const next = () => updateCurrentStep(stepState.next)

export const back = () => updateCurrentStep(stepState.back)

export const set = (step: number) => updateCurrentStep(() => stepState.set(step))
```

WDYT?

src/lib/components/WizardModal.svelte
AntonioVentilii: is fired only once, when `stepState` changes (it is derived based on input params `steps`). There is no conflict, as per what I tested. The functions `next`, `back` and `set` do not trigger the `$effect`, and the `$effect` is triggered only when `stepState` is effectively re-defined

peterpeterparker: If the `steps` are updated, `stepState` changes and the effect is triggered?

src/lib/components/WizardModal.svelte
peterpeterparker: Sounds. I would rename `updateCurrentStep` and `set: () => ..` though. The former because it doesn't "just" update current step anymore and the later because there is also `stepState.set` but `set` can actually be back or next. You know what I mean?

src/lib/components/WizardModal.svelte
AntonioVentilii: correct: `steps` changes --> `stepState` is re-defined from scratch --> `currentStep` is updated

src/lib/components/WizardModal.svelte
AntonioVentilii: I tried a few variations but nothing convinced me. Have a look at the lates, WDYT? Suggestions?

AntonioVentilii: Comment included!

```ts
const run = (transition: () => WizardStepsState<T>) => {
    stepState = transition();
    ({ currentStep } = stepState);
  };
```
  
 or
 
```ts
 const runStepTransition = (run: () => WizardStepsState<T>) => {
    stepState = run();
    ({ currentStep } = stepState);
  };
```

?

peterpeterparker (approved):
LGTM, thx

src/tests/lib/components/WizardModal.spec.ts
```diff
@@ -63,4 +64,60 @@ describe("WizardModal", () => {
...
+  it("should trigger the next step when modal.next is called", async () => {
+    const { getByTestId } = render(WizardModalTest);
+
+    const nextButton = getByTestId("next-button");
+
+    expect(nextButton).toBeInTheDocument();
+
+    await fireEvent.click(nextButton);
+
+    await waitFor(() => {
```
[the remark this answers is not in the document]
peterpeterparker: Ahhhh it's because next and back a not exposed as arrow function.

```
// Is - does not work
public next(): WizardStepsState<T> {

// Would probably work
public next = (): WizardStepsState<T> => {
```

But for sure we do not change that, it's fine. Thanks for fixing!

PR #700, opened 2025-07-28
feat: failed step in ProgressSteps component

roman-nazaruk:
# Motivation

We need to introduce a new step in the progress flow to represent failed states. This allows better feedback to the user when a step in the process doesn't complete successfully.

# Changes

- Added support for a new `failed` state in `ProgressSteps`
- Updated styles to show the failed step clearly (color, border, icon, label)

# Screenshots
image


peterpeterparker (commented):
Code looks good. The icon is maybe few pixels too small? WDYT?

Capture d’écran 2025-07-28 à 11 38 36
Capture d’écran 2025-07-28 à 11 38 40

src/lib/icons/IconCheckCircleFill.svelte
```diff
@@ -13,7 +13,7 @@
   width={size}
   height={size}
   viewBox="0 0 20 20"
-  fill="currentColor"
+  fill="var(--icon-check-circle-background, transparent)"
```
peterpeterparker, line 16: Thinking at it now, maybe we should rename it `--icon-check-circle-fill-background` to avoid side effect?

peterpeterparker (approved):
LGTM, thx

Just added one final comment that crossed my mind when reviewing it one last time.

src/lib/icons/IconCheckCircleFill.svelte
roman-nazaruk: Sure, good idea. Updated

PR #701, opened 2025-07-30
fix: reverted progressSteps component and added new IconCloseCircle for failed step

roman-nazaruk:
# Motivation

Refer to [PR #700:](https://github.com/dfinity/gix-components/pull/700)
The changes introduced in this PR caused some style issues. It was decided to revert to the previous implementation and introduce a new IconCloseCircle that matches the size of IconCheckCircle.

# Changes
- Reverted styles and IconCheckCircle to the previous version
- Added new IconCloseCircle with matching dimensions

# Screenshots

image

image

### DIFF
--- a/src/lib/components/ProgressSteps.svelte
+++ b/src/lib/components/ProgressSteps.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
   import Spinner from "$lib/components/Spinner.svelte";
-  import IconCheckCircleFill from "$lib/icons/IconCheckCircleFill.svelte";
-  import IconCloseCircleFill from "$lib/icons/IconCloseCircleFill.svelte";
+  import IconCheckCircle from "$lib/icons/IconCheckCircle.svelte";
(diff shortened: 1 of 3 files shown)


src/lib/components/ProgressSteps.svelte
```diff
@@ -1,7 +1,7 @@
 <script lang="ts">
   import Spinner from "$lib/components/Spinner.svelte";
-  import IconCheckCircleFill from "$lib/icons/IconCheckCircleFill.svelte";
-  import IconCloseCircleFill from "$lib/icons/IconCloseCircleFill.svelte";
+  import IconCheckCircle from "$lib/icons/IconCheckCircle.svelte";
+  import IconCloseCircle from "$lib/icons/IconCloseCircle.svelte";
```
peterpeterparker, line 4: You need to list the new icon in `icons.ts` otherwise it won't be exposed to the consumer of the lib.

roman-nazaruk: Added. Thanks

PR #706, opened 2025-08-25
chore(license): removes modified boilerplate from license file

jwndlng:
This PR removes the modified appendix from the LICENSE file to ensure full compliance with the Apache 2.0 license, which should remain in its original, unmodified form. This change aligns our repositories with standard practices for license documentation.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -174,28 +174,3 @@
       of your accepting any such warranty or additional liability.
 
    END OF TERMS AND CONDITIONS
-
-   APPENDIX: How to apply the Apache License to your work.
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
Thanks.

You've got other gix repo that requires similar changes under radar or shall we proceed with similar change?

jwndlng:
Hey David

I am currently going through all dfinity repos, so there is no action required from your end. 

> Also I assume this does not require an urgent release to npm, is that a safe assumption?
That's right, no release needed.

PR #707, opened 2025-08-28
fix(markdown): preserve SVG tags in code blocks

yhabib:
# Motivation

The `Markdown` component converts markdown into HTML. For security reasons, the input is sanitized before transformation.

The initial implementation was introduced [here](https://github.com/dfinity/nns-dapp/pull/3399) and acknowledged this edge case.
> It's not possible to use the HTML renderer since the SVG contains multiple tags. 
> One edge case remains unaddressed: if the SVG is inside the `<code>` tag, it will be rendered with &lt; and &gt; instead of "<" and ">."

We need to address this use case now, as there is a mismatch between how the proposal is rendered and how it should be rendered:  
* https://nns.ic0.app/proposal/?u=qoctq-giaaa-aaaaa-aaaea-cai&proposal=138188  
* https://dashboard.internetcomputer.org/proposal/138188

# Changes

* Do not escape svg's inside code blocks.

# Screenshots

Screenshot 2025-08-27 at 15 42 30

### DIFF
--- a/src/lib/utils/markdown.utils.ts
+++ b/src/lib/utils/markdown.utils.ts
@@ -127,7 +127,6 @@ const proposalSummaryRenderer = (marked: Marked): Renderer => {
 export const markdownToHTML = async (text: string): Promise<string> => {
   // Replace the SVG elements in the HTML with their escaped versions to improve security.
   // It's not possible to do it with html renderer because the svg consists of multiple tags.
-  // One edge case is not covered: if the svg is inside the <code> tag, it will be rendered as with &lt; & &gt; instead of "<" & ">"
   const escapedText = escapeSvgs(text);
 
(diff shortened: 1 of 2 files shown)


src/lib/utils/markdown.utils.ts
```diff
@@ -47,8 +47,36 @@ export const imageToLinkRenderer = (
...
+  }
+
+  // Match inline code (`...`)
+  const inlineCodeRegex = /`[^`\n]+`/g;
+  while ((match = inlineCodeRegex.exec(html)) !== null) {
+    codeBlocks.push({ start: match.index, end: match.index + match[0].length });
+  }
+
+  // Sort code blocks by start position
+  codeBlocks.sort((a, b) => a.start - b.start);
```
mstrasinskis, line 69: Is this step actually an improvement, given that no binary search is used?

mstrasinskis (approved):
Thank you!

PR #709, opened 2025-09-09
fix: Flaky E2E test for Progress steps

AntonioVentilii:
# Motivation

A new step was included in the Progress step component showcase, but that means that we should wait slightly more for the animation to finish before taking the snapshot, to be sure and avoid flakiness.


peterpeterparker (approved):
LGTM, thx

PR #818, opened 2026-05-12
feat(popover): auto-flip horizontal placement when the preferred side overflows the viewport

AntonioVentilii:
## Motivation

`Popover` pins itself to one anchor edge based on `direction` with no collision detection, so a trigger near a viewport edge can produce a panel that overflows toward that edge. Switching to the opposite `direction` just moves the problem to the other end. Even when the panel is allowed to fall back to whichever side has more room, the panel still gets compressed (or overflows) the moment its natural width does not fit anchored to that side — vertical placement already handles this more gracefully via `max-height`.

## Changes

The PR builds up the horizontal equivalent of the existing vertical containment, in layered steps:

1. **Honor the preferred side.** `direction` is treated as a preference; the new `pickPopoverDirection` keeps it when the panel fits.
2. **Flip on overflow.** If the preferred side overflows, flip to the opposite side when that one fits.
3. **Shift along the viewport.** When neither side has room for the panel's natural width, `computePopoverPlacement` slides the panel along the viewport (with a `--padding` margin) so it stays inside at its natural width. This mirrors how the existing `max-height` already keeps the panel inside vertically.
4. **Clamp as a last resort.** Only when the panel is wider than the usable viewport (`viewport - 2 * padding`), the existing `max-width` rule (gated by a new `.placed` class) compresses the panel. The `.placed` class is only applied after the first measurement so the auto-flip decision sees the panel's natural width.

The placement is recomputed on resize, on `anchor` / `visible` / `direction` changes, and whenever the measured panel width changes (via `ResizeObserver`).


src/lib/components/Popover.svelte
```diff
@@ -14,24 +15,115 @@
...
+    if (typeof window === "undefined") {
+      return;
+    }
+
+    if (!anchor) {
+      popoverTop = 0;
+      popoverLeft = 0;
+      popoverRight = 0;
+      effectiveDirection = direction;
+      return;
```
AntonioVentilii, line 57:
```suggestion

      return;
```

yhabib (approved):

